PR #10, opened 2015-05-28
restore proper version of spipe*/Makefile

gperciva:
alternatively, you might want to ignore these files in git; there isn't much point tracking auto-generated files.


cperciva:
Huh, this is weird.  Did you generate these via 'make Makefiles'?  I wonder if your make is handling variable expansion differently from FreeBSD's make.

But I do want them in the tree, because if people check out this tree on linux I want them to have the makefiles they need -- they might not be able to (easily) regenerate them.

gperciva:
I did
    bmake publish VERSION=mine
and those files were changed.  I figured that ${PROG}.1 was a more "git
master"-ish thing to have.

(incidently, I created this comment by hitting "reply" in my email
client)

- Graham

cperciva:
Ok, something weird going on here.  Can you:

    cd spipe && bmake -f Makefile.BSD -V MAN1

and tell me what it returns?

gperciva:
$ cd spipe && bmake -f Makefile.BSD -V MAN1
${PROG}.1

gperciva:
if I change Makefile.BSD to:

    # Program name.
    PROG    =       spipe
    MAN1    :=      ${PROG}.1

then I get:

    gperciva@gin: ~/src/spiped/spipe (master)
    $ bmake -f Makefile.BSD -V MAN1
    spipe.1

cperciva:
Interesting... yes, definitely different behaviour between FreeBSD and NetBSD make -V.  I wonder if this is a known issue?

The := option looks like it's going to be the right way of handling this though.  I'll take care of it in a bit.

cperciva:
Can you check out the latest master, run

    bmake Makefiles
    git diff

and tell me if anything changed?  I *think* my recent commit works around the NetBSD/FreeBSD make incompatibility.

gperciva:
Done, and no changes.

cperciva:
Perfect, we don't need this then.

PR #12, opened 2015-05-28
Version numbers 2

gperciva:


gperciva:
The best two options I see so far are:

1) always write a backup file with

    sed -i.bak -e '...' filename

2) use Perl.  This seems really heavy, though.  :(

cperciva:
Seriously?  *grumble*

Maybe we need to go back to sed + mv after all.

And yes, I'd like anyone to be able to run 'make publish' -- unlike 'make all' I'm willing to demand that they have BSD make since they can invoke it as bmake, but I can't demand that sed is BSD sed.

gperciva:
if we *do* go back to sed+mv, can we assume that they have a /tmp?  Apparently `mktemp` is not in POSIX, and is missing from Solaris 9 at least.

or do we just write to "our-filename-sed-repace.tmp" in the ${PKG}-{VERSION}/ directory?

cperciva:
No harm in writing into the staging directory.  Let's make it

    sed -e ... < file > file.new
    mv file.new file

gperciva:
done, pushed to a separate branch.

PR #23, opened 2015-08-02
Fix 'docker stop' issue

TimWolla:
Docker runs processes inside a container with an isolated process tree.
Usually your daemon is the only process inside this container. This means
that it receives PID 1, being effectively init. This leads to the default
signal handlers of SIGTERM and SIGINT being ignored, which in turn causes
very slow stops of a container running spiped (it waits for the SIGKILL).

This commit explicitely bind a signal handler to SIGTERM and SIGINT if
spiped detects it is PID 1.

--------
Example:
```
[timwolla@/d/w/c/spiped (signal_handler) [1]]
docker run -v (pwd)/keyfile:/spiped/key:ro -p 8025:8025 timwolla/spiped-test -d -s '[0.0.0.0]:8025' -t '[127.0.0.1]:25'
spiped: WARNING: Applying workaround for Docker signal-handling bug
^C⏎ 
[timwolla@/d/w/c/spiped (signal_handler)]
./spiped/spiped -F -d -s '[0.0.0.0]:8025' -t '[127.0.0.1]:25' -k keyfile
^C⏎ 
[timwolla@/d/w/c/spiped (signal_handler) [130]]
```

### DIFF
--- a/spiped/main.c
+++ b/spiped/main.c
@@ -1,4 +1,5 @@
 #include <inttypes.h>
+#include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -25,6 +26,18 @@ usage(void)
 	    "    [-p <pidfile>] [-r <rtime> | -R]\n"
 	    "       spiped -v\n");
 	exit(1);
+}
+
+/*
+ * Signal handler for SIGTERM and SIGINT in case we are running
+ * inside a container as PID 1.
+ */
+static void
+diediedie_handler(int signo)
+{
+
+	(void)signo; /* UNUSED */
+	_exit(0);
 }
 
 /* Simplify error-handling in command-line parse loop. */
@@ -195,6 +208,20 @@ main(int argc, char * argv[])
 		if (asprintf(&opt_p, "%s.pid", opt_s) == -1) {
 			warnp("asprintf");
 			exit(1);
+		}
+	}
+
+	/* Check whether we are running as init (e.g., inside a container). */
+	if (getpid() == 1) {
+		/* https://github.com/docker/docker/issues/7086 */
+		warn0("WARNING: Applying workaround for Docker signal-handling bug");
+
+		/* Bind an explicit signal handler for SIGTERM and SIGINT. */
+		if (signal(SIGTERM, diediedie_handler) == SIG_ERR) {
+			warnp("Failed to bind SIGTERM signal handler");
+		}
+		if (signal(SIGINT, diediedie_handler) == SIG_ERR) {
+			warnp("Failed to bind SIGINT signal handler");
 		}
 	}
 


spiped/main.c
```diff
@@ -27,6 +28,12 @@ usage(void)
 	exit(1);
 }
 
+static void
+signal_handler(int signo)
+{
+	exit(0);
```
cperciva: This needs to be _exit since the 'exit' function can't be used from inside a signal handler: http://pubs.opengroup.org/onlinepubs/9699919799/

spiped/main.c
```diff
@@ -27,6 +28,12 @@ usage(void)
 	exit(1);
 }
 
+static void
```
cperciva: Add a comment here explaining what this function is for.

spiped/main.c
```diff
@@ -27,6 +28,12 @@ usage(void)
...
+signal_handler(int signo)
```
cperciva: I'd rename this to "diediedie_handler" to make the intention clearer.  "signal_handler" is a bit vague...

spiped/main.c
cperciva: Style requires a blank line separating the (in this case empty) local variable declarations from the code in the function.

cperciva: Unused variables need to be explicitly designated as such:

        (void)signo; /* UNUSED */

In addition to making the programmer intention clear, it prevents compiler (and static analysis tool) warnings.

spiped/main.c
```diff
@@ -198,6 +205,19 @@ main(int argc, char * argv[])
 		}
 	}
 
+	/* Check whether we are init (i.e. running inside a container). */
```
cperciva: We aren't init.  But we might be *running as* init.  I'd suggest:

    /* Check whether we are running as init (e.g., inside a container). */

spiped/main.c
```diff
@@ -198,6 +205,19 @@ main(int argc, char * argv[])
...
+	if (getpid() == 1) {
+		/* https://github.com/docker/docker/issues/7086 */
+		warn0("WARNING: Applying workaround for Docker signal-handling bug");
+		/* Bind explicit signal handler for SIGTERM and SIGINT. */
```
cperciva: Blank line between code and following comment.

cperciva: I'm not a fan of the grammar here either.  s/Bind explicit/Bind an explicit/ perhaps?

spiped/main.c
TimWolla: > Unused variables need to be explicitly designated as such:

Literally that line? Above or below the `_exit`? Separated with a blank line?

cperciva: After the blank line which separates variables from code.  Before _exit(0) since otherwise it would be dead code.

So the function should look like this:

    {
        [blank line]
        (void)signo; /* UNUSED */
        _exit(0);
    }

TimWolla:
@cperciva I just pushed a commit that should fix the issues. Thanks for the review, so far!

cperciva:
Thanks, looks good.  Uhh, dumb git question, is there some way you can squish these two commits into one before I merge?  I don't think there's any need to have the first attempt separately in spiped's git history.

TimWolla:
@cperciva Sure, I just squashed the commits.

To explain how it's done:

1. Make sure you are on your feature branch (`signal_handler`) in my case.
2. `git rebase -i origin/master`
3. Decide what to to with each commit. In my case `pick` the first and `fixup` for the second.
4. `git push --force` into the remote repository.

cperciva:
Thanks!  I knew that the squishing involved `git rebase` but I wasn't sure exactly how that interacted with pull requests and github.

TimWolla:
> but I wasn't sure exactly how that interacted with pull requests and github.

Luckily pretty much as expected :smile: 

Thanks for merging, help and the IRC conversation!

Issue #26, opened 2015-10-21
1.5.0.tar.gz from GitHub isn't buildable with GNU Make - the one from tarsnap.com is fine


BonkaBonka:
It appears to require BSD make.  GNU make reports:

    $ make
    Makefile:5: *** missing separator.  Stop.

TimWolla:
[Updating build code and releasing](https://github.com/Tarsnap/spiped#updating-build-code-and-releasing) explains that the repository contains BSD Makefiles. The 1.5.0.tar.gz simply is an automated export of the tree in the git tag, without any further processing.

cperciva:
What Tim said.  Github's repository snapshot tarballs are not release tarballs.

gperciva:
This seems incredibly misleading [of github].  I uploaded spiped-1.5.0.tgz (currently tagged as draft): https://github.com/Tarsnap/spiped/releases

Shall I un-draft that, and upload other tarballs?  or maybe only have the latest release?

gperciva:
oops, apparently "draft" is only viewable by me.  I tried undrafting it, but github automatically added the unhelpful zip and git-tarball.  Sorry for the noise.

I'm going to play in one of my local repositories (one which doesn't have 51 stars) to figure out how to stop github from unhelpfully adding snapshots.

gperciva:
Looks like a release absolutely needs a git tag, and there's no way to make it stop adding tag-based snapshots.  Technically, we could create a branch for each release, add the generated POSIX makefiles, and tag that as the release.

But (after discussing briefly with Colin offline), I think it's simpler just to point people at tarsnap.com for stable releases.

cperciva:
Nobody on twitter seems to know how to disable the automatic "release" tarballs either.  Sigh, why does github have to be so broken?

Anyway, not much we can do this except for adding the warning (already done) so I'm going to close this again.  Feel free to follow up if you have any idea how to turn off this github mis-feature.

PR #27, opened 2015-10-22
README.md: add warning to download from tarsnap.com

gperciva:
The github "releases" are merely snapshots of the git repository, which uses
BSD Makefiles.

This adds emoji to the README.md, because spiped is a serious project and serious projects need emoji.  (they make sense, I swear!)

This addresses the root cause of #26


gperciva:
You can view the new github page in all its emoji glory:
https://github.com/Tarsnap/spiped/tree/readme

(if I was writing documentation, I'd use a box with a different-colored background and a red "warning" title, but it seemed easier to do it this way, and it still makes sense when you read the file as pure text)

PR #29, opened 2015-12-20
Compatibility with /dev/stderr and su and linux

gperciva:
This was a fun one!  TL;DR: my OS doesn't update the ownership of /dev/stderr
when su'ing into a new account, so attempting to redirect to /dev/stderr fails
with a "permission denied".

My main user (the one running X11) is gperciva.  I opened an xterm, did
$ su td
    (for "tarsnap development"), and then
td@gin: ~/src/spiped/libcperciva/POSIX (stderr-compatibility)
$ ls -lL /dev/stderr
crw--w---- 1 gperciva tty 136, 2 Dec 20 11:43 /dev/stderr

According to at least one comment on an internet forum (what could go
wrong?!), the shell is *supposed* to treat /dev/stderr as a special case (and
internally use 1>&2), but if that file exists, then it attempts to write
directly to it.  Which fails, since ownership of /dev/stderr is not updated
when using su.

Long description here:
http://unix.stackexchange.com/questions/38538/bash-dev-stderr-permission-denied
(I'm using dash instead of bash, but the problem is the same)

### DIFF
--- a/libcperciva/POSIX/posix-cflags.sh
+++ b/libcperciva/POSIX/posix-cflags.sh
@@ -1,10 +1,10 @@
 # Should be sourced by `command -p sh posix-cflags.sh` from within a Makefile
 if ! ${CC} -D_POSIX_C_SOURCE=200809L posix-msg_nosignal.c 2>/dev/null; then
 	printf %s "-DPOSIXFAIL_MSG_NOSIGNAL "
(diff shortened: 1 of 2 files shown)


cperciva:
Wow.  Just... wow.  How is it possible for open("/dev/stderr") to not be identical to dup(2)?

PR #33, opened 2016-01-27
add error path to spiped

gperciva:
first draft of adding error paths to spiped.

### DIFF
--- a/spiped/dispatch.c
+++ b/spiped/dispatch.c
@@ -189,24 +189,26 @@ dispatch_accept(int s, const char * tgt, double rtime, struct sock_addr ** sas,
 	/* If address re-resolution is enabled... */
 	if (rtime > 0.0) {
 		/* Launch an address resolution thread. */
-		A->T = dnsthread_spawn();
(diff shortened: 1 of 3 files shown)


spiped/main.c
```diff
@@ -302,4 +303,18 @@ main(int argc, char * argv[])
 	 * If we could reach this point, we would free memory, close sockets,
```
TimWolla: There is a `NOTREACHED` above, it should be removed, no?

cperciva: The `/* NOTREACHED */` is correct since we never fall out of the loop; all we ever do is jump to error-handling.

spiped/dispatch.c
```diff
@@ -27,6 +27,7 @@ struct accept_state {
 	size_t nconn_max;
 	double timeo;
 	void * accept_cookie;
+	void * event_cookie;
```
cperciva: `event_cookie` is a bit vague, since every event (network, timer, immediate) we can schedule provides a cookie.  How about `dnstimer_cookie`?

gperciva:
Updated with `dnstimer_cookie`.

PR #34, opened 2016-02-07
Add proto_conn_shutdown() and error path to spipe

gperciva:
The error path only applies to the first half of spipe main.c; cleaning up the
sockets and pthreads is deferred to a later commit.

### DIFF
--- a/proto/proto_conn.c
+++ b/proto/proto_conn.c
@@ -197,9 +197,10 @@ dropconn(struct conn_state * C)
  * only if ${nokeepalive} is zero.  Drop the connection if the handshake or
  * connecting to the target takes more than ${timeo} seconds.  When the
  * connection is dropped, invoke ${callback_dead}(${cookie}).  Free ${sas}
- * once it is no longer needed.
+ * once it is no longer needed.  Returns a cookie which can be passed to
+ * proto_conn_shutdown.
  */
-int
+void *
 proto_conn_create(int s, struct sock_addr ** sas, int decr, int nofps,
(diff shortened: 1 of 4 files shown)


cperciva:
`proto_conn_create` is also used by spiped.  Have you tested that?

gperciva:
Oops, I tested that it compiled, but not that it worked correctly.  Revising this PR.

gperciva:
PR updated, but could use more testing.  Following up in private email, unless you want to play Knuth and accept it because it's been proven to be correct.

proto/proto_conn.c
```diff
@@ -392,3 +393,19 @@ callback_pipestatus(void * cookie)
...
+void
+proto_conn_shutdown(void * conn_cookie)
+{
+	struct conn_state * C = conn_cookie;
+
+	if (C->connect_cookie != NULL)
+		network_connect_cancel(C->connect_cookie);
+	if (C->connect_timeout_cookie != NULL)
+		events_timer_cancel(C->connect_timeout_cookie);
+	free(C);
```
cperciva: What about `pipe_f` and `pipe_r` ?

spiped/dispatch.c
```diff
@@ -28,6 +28,7 @@ struct accept_state {
 	double timeo;
 	void * accept_cookie;
 	void * dnstimer_cookie;
+	void * conn_cookie;
```
cperciva: spiped might have lots of connections.  Why are you holding a pointer to just one?

gperciva:
This needs a lot of reworking.

PR #36, opened 2016-02-07
Run spiped only once with -1

gperciva:
I put the man-page documenting of `-1` in a separate commit since you might tell me it should be undocumented.  My vote (which has no value) is that we should still document it, because I won't be able to remember it otherwise, and it's not like anybody can do anything nasty if they know about `-1`.  (and even if they could, they could discover it from the source code anyway).

Since `spipe` can quit with `^D`, I didn't add `-1` there, but you might want me to do so anyway for consistency.

### DIFF
--- a/spiped/dispatch.c
+++ b/spiped/dispatch.c
@@ -158,7 +158,7 @@ err0:
  * the addresses ${sas}.  If ${decr} is 0, encrypt the outgoing connections; if
  * ${decr} is non-zero, decrypt the incoming connections.  Don't accept more
  * than ${nconn_max} connections.  If ${nofps} is non-zero, don't use perfect
- * forward secrecy.  If {$requirefps} is non-zero, require that both ends use
+ * forward secrecy.  If ${requirefps} is non-zero, require that both ends use
  * perfect forward secrecy.  Enable transport layer keep-alives (if applicable)
  * if and only if ${nokeepalive} is zero.  Drop connections if the handshake or
  * connecting to the target takes more than ${timeo} seconds.  Returns a
--- a/spiped/dispatch.c
+++ b/spiped/dispatch.c
@@ -22,6 +22,7 @@ struct accept_state {
 	int nofps;
 	int requirefps;
 	int nokeepalive;
+	int * conndone;
 	const struct proto_secret * K;
 	size_t nconn;
 	size_t nconn_max;
@@ -98,6 +99,10 @@ callback_conndied(void * cookie)
 	/* We've lost a connection. */
 	A->nconn -= 1;
 
+	/* If requested to do so, indicate that a connection closed. */
+	if (A->conndone != NULL)
+		*A->conndone = 1;
+
 	/* Maybe accept more connections. */
 	return (doaccept(A));
 }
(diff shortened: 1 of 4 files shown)


cperciva:
Err... I thought we were going to make `-1` handling just a single *connection*?  Which may involve many calls to `events_run`...

cperciva:
(And spipe doesn't need this since it already only handles one connection.)

gperciva:
Updated according to our discussion this afternoon.  If this is accepted, I'll revise spipe along the same lines for consistency.

spiped/spiped.1
```diff
@@ -115,5 +116,8 @@ Disable target address re-resolution.
 .TP
 .B \-v
 Print version number.
+.TP
+.B \-1
+Only accept one event, then exit.  Used for internal testing.
```
cperciva: Needs to be updated.  "Exit after a connection closes." ?

spiped/main.c
```diff
@@ -285,20 +292,28 @@ main(int argc, char * argv[])
...
-			goto err6;
-		}
-	} while (1);
+	if (events_spin(&conndone)) {
+		warnp("Error running event loop");
+		goto err6;
+	}
+
+	/* Clean up. */
+	dispatch_shutdown(dispatch_cookie);
```
cperciva: In non-error-handling paths I usually have comments explaining what the things are which we're freeing.  e.g., "Stop accepting connections and shut down the dispatcher", "Shut down the events system", "Free arrays of resolved addresses", etc.

spiped/dispatch.c
```diff
@@ -98,6 +99,10 @@ callback_conndied(void * cookie)
 	/* We've lost a connection. */
 	A->nconn -= 1;
 
+	/* If only one connection was desired, quit event loop. */
```
cperciva: POV: The dispatcher doesn't know anything about users or command lines or event loops; all it knows is what the call to dispatch_accept told it, i.e., that sometimes we're supposed to set this to a non-zero value when a connection closes.

cperciva:
Are you keeping test scripts somewhere?  Maybe worth adding them to the tree?

Also, just a reminder: You need a test script which creates two connections through `spiped -1` and then closes one.  Right now this will result in the second connection not being cleaned up.

gperciva:
I'll add test scripts in a later PR; I've it written down on my TODO list now.

I know (or suspected, at least) that spiped would be leaking some stuff; my goal in this PR was to get spiped into a state that I could easily test for those, then do memory cleanup in a separate PR.

gperciva:
Updated.  As per our offline discussion, I've left my updates as separate commits; if they look ok, I'll rebase the patch so that the history looks nicer.

spiped/main.c
```diff
@@ -285,20 +292,36 @@ main(int argc, char * argv[])
...
 	/* Start accepting connections. */
 	if ((dispatch_cookie = dispatch_accept(s, opt_t, opt_R ? 0.0 : opt_r,
-	    sas_t, opt_d, opt_f, opt_g, opt_j, K, opt_n, opt_o)) == NULL) {
+	    sas_t, opt_d, opt_f, opt_g, opt_j, K, opt_n, opt_o,
+	    opt_1 ? &conndone : NULL)) == NULL) {
 		warnp("Failed to initialize connection acceptor");
 		goto err5;
 	}
 
 	/* Infinite loop handling events. */
```
cperciva: `/* Loop until an error occurs, or a connection closes if running with -1. */`

cperciva:
Looks good aside from one comment tweak.  Clean up the commit history and then I'll merge.

gperciva:
Updated with comment change and cleaned up history.  The proposed comment doesn't fit onto one line, so I expanded it slightly to make use of the extra space:

```
/*
 * Loop until an error occurs, or a connection closes if the
 * command-line argument -1 was given.
 */
```

PR #37, opened 2016-02-16
libcperciva: initialize sockerr=0

gperciva:
On FreeBSD 10.2-RELEASE-p9, valgrind-freebsd-3.10.0 claims that with a
successful getsockopt(), &sockerr is not written to, so:

    if (sockerr != 0)

was reading uninitialized memory.

### DIFF
--- a/libcperciva/network/network_connect.c
+++ b/libcperciva/network/network_connect.c
@@ -65,7 +65,7 @@ static int
 callback_connect(void * cookie)
 {
 	struct connect_cookie * C = cookie;
-	int sockerr;
+	int sockerr = 0;
 	socklen_t sockerrlen = sizeof(int);
 
 	/* Stop waiting for the timer callback. */


gperciva:
I'm not quite believing this, especially since if I set sockerr to something non-zero before calling getsockopt(), sockerr gets set to 0 after getsockopt().  But I figure I might as well put this code out there so we can discuss it.

gperciva:
In case it's relevant, valgrind-freebsd does not complain about `sockerrlen`, which is passed into `getsockopt` in the same way.

gperciva:
EDIT to previous comment: no wait, my bad.  `sockerrlen` is assigned a value before calling `getsockopt`.

gperciva:
I just tested valgrind-freebsd on tarsnap, and it reports a lot of leaks for silly things like `getpwuid`.  So on second thought, I'm inclined to just ignore this as a false positive and close the PR without merging.

cperciva:
There is actually a leak in FreeBSD libc's `getpwuid`.  Not a serious one, but a "data structures are initialized and not cleaned up on exit".  Bryan Drewery was looking at cleaning up some of these (https://github.com/bdrewery/freebsd/tree/libc_freeres).

That's completely separate from the `getsockopt` issue though.  Valgrind runs in userland and can track what libc does; but it has no idea what the kernel does and has no understanding of whether values passed into the kernel are being read or not.

cperciva:
This is definitely a valgrind false positive though, and I'd rather leave the false positive around in case it helps to get valgrind fixed.  I'm guessing it has some modelling of what the kernel should do with parameters but that modelling is insufficient on FreeBSD.

PR #38, opened 2016-03-08
Add POSIX test script for spiped

gperciva:


tests/spiped/test_spiped.sh
```diff
@@ -0,0 +1,254 @@
...
+	if $system_spiped_binary -1 2>&1 >/dev/null | grep -q "invalid option"; then
+		# disable test
+		system_spiped_binary=""
+	fi
+fi
+
+# Check for required commands
+if ! command -v ncat >/dev/null 2>&1; then
+	echo "ncat not detected; it is part of the nmap suite"
+	echo "	(it is required for multiple sockets on the same port"
```
cperciva: missing `)` ?

tests/spiped/test_spiped.sh
```diff
@@ -0,0 +1,254 @@
...
+		$system_spiped_binary -d \
+		-s 127.0.0.1:$mid_port -t 127.0.0.1:$dst_port \
+		-k keys-blank.txt -F -1 \
+	; echo $? >> $out/$basename-d.ret ) &
+	sleep $sleep_valgrind_start
+}
+
+
+####################################################
+test_single () {
```
cperciva: Can we have some more useful test names or descriptions?

This test seems to be confirming that if spiped tries to connect to something which doesn't speak the spiped protocol, then spiped will drop the connection?

gperciva:
PR updated to clarify goals of tests.

gperciva:
@cperciva wrote:
> ... so the connection-closing is actually going to happen even if we don't try to close the connection ourselves.
>
> Maybe this should be two tests, one with spiped talking to ncat and timing out, and a second one with spiped talking to spiped, so that we can confirm that the cleanup happens whether the handshake is completed or not?

I'm not certain I follow.  Isn't the case of "spiped talking to spiped but sending no data" covered by `test_connection_open_close_double` ?

I added another test of opening a connection to a single spiped server, but not sending anything; this checks that spiped will time out.

cperciva:
Err, yeah, I said that wrong.  I meant two tests with one closing the connection deliberately and the other having the timeout happen.

gperciva:
updated PR

cperciva:
Can you hook this into the Travis-CI testing?

gperciva:
ok, this will take some travis-ci magic to install nmap on osx with homebrew, and to only use apt-get on linux.  Looking into it.

gperciva:
Despite the tests "passing", there's problems with the test infrastructure, particularly on OSX.  I'll need to come back to this again later.

gperciva:
This is blocked on a problem with the new `-1` option.  I'm working on a proper bug report (or fix) for that.

gperciva:
Simplifying it a bit; will re-open later.

gperciva:
I have the valgrind tests ready as well, but they don't pass at the moment (I'm working on that).

PR #39, opened 2016-03-17
Cppflags

gperciva:
This was adapted from @bgilbert's https://github.com/Tarsnap/tarsnap/pull/147 .


cperciva:
1. The -I. isn't redundant here.  It's only redundant in automake because autotools adds it automatically.

2. You know this is only used by `Makefile.BSD` anyway, right?  The POSIX Makefiles don't use this.

gperciva:
Further development of makefiles is taking place in https://github.com/Tarsnap/spiped/pull/43.

PR #41, opened 2016-04-04
Add -c <chmod> support to spiped to set the permissions for created Unix sockets.

wyattoday:
A justification for this addition: we use spiped at boot time on a number of our servers. And we need the Unix socket spiped creates to be available immediately to other processes. There's no good way to do this on our systems which use the "Upstart" init system (short of creating another processes that waits for spiped to create the socket, then altering the permissions of the socket).

We're already using these modifications in production.

### DIFF
--- a/libcperciva/util/sock.c
+++ b/libcperciva/util/sock.c
@@ -328,11 +328,11 @@ sock_listener(const struct sock_addr * sa, mode_t sm)
 	}
 
 	/* Set the Unix socket's permission. */
-	if (sa->ai_family == AF_UNIX && sm != -1) {
-		if (chmod(((struct sockaddr_un *)sa->name)->sun_path, sm)) {
-			warnp("Error calling chmod on Unix socket");
-			goto err1;
-		}
+	if (sa->ai_family == AF_UNIX && sm != -1
+			&& chmod(((struct sockaddr_un *)sa->name)->sun_path, sm)) {
+
+		warnp("Error calling chmod on Unix socket");
+		goto err1;
 	}
 
 	/* Mark the socket as listening. */
--- a/libcperciva/util/sock.c
+++ b/libcperciva/util/sock.c
@@ -327,9 +327,12 @@ sock_listener(const struct sock_addr * sa, mode_t sm)
 		goto err1;
 	}
 
-	/* Set the Unix socket's permission */
+	/* Set the Unix socket's permission. */
 	if (sa->ai_family == AF_UNIX && sm != -1) {
-		chmod(((struct sockaddr_un *)sa->name)->sun_path, sm);
+		if (chmod(((struct sockaddr_un *)sa->name)->sun_path, sm)) {
+			warnp("Error calling chmod on Unix socket");
+			goto err1;
+		}
 	}
 
 	/* Mark the socket as listening. */
--- a/libcperciva/util/sock.c
+++ b/libcperciva/util/sock.c
@@ -300,10 +300,11 @@ err0:
 /**
  * sock_listener(sa):
  * Create a socket, set SO_REUSEADDR, bind it to the socket address ${sa},
- * mark it for listening, and mark it as non-blocking.
+ * set the permission (if not -1), mark it for listening,
+ * and mark it as non-blocking.
  */
 int
-sock_listener(const struct sock_addr * sa)
+sock_listener(const struct sock_addr * sa, mode_t sm)
 {
 	int s;
 	int val = 1;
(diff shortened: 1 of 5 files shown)


libcperciva/util/sock.c
```diff
@@ -300,10 +300,11 @@ sock_resolve(const char * addr)
 /**
  * sock_listener(sa):
```
gperciva: add the `sm` argument here

libcperciva/util/sock.c
```diff
@@ -300,10 +300,11 @@ sock_resolve(const char * addr)
...
  * Create a socket, set SO_REUSEADDR, bind it to the socket address ${sa},
- * mark it for listening, and mark it as non-blocking.
+ * set the permission (if not -1), mark it for listening,
```
gperciva: I think you mean "(if ${sm} is not -1)".

spiped/README
```diff
@@ -8,8 +8,8 @@ usage: spiped {-e | -d} -s <source socket> -t <target socket> -k <key file>
 
 Options:
     -c <source chmod>
-	The access permissions of the source socket (if it's a Unix socket).
```
gperciva: I think the original version was correct -- these lines should begin with a single `TAB` character.

spiped/main.c
```diff
@@ -25,6 +25,7 @@ usage(void)
 	    "    [-DFj] [-f | -g] [-n <max # connections>] "
 	    "[-o <connection timeout>]\n"
 	    "    [-p <pidfile>] [-r <rtime> | -R] [-1]\n"
+		"    [-c <source chmod>]\n"
```
gperciva: whitespace here should be `TAB    ` (tab and then 4 spaces).

spiped/main.c
```diff
@@ -82,6 +84,20 @@ main(int argc, char * argv[])
 	/* Parse the command line. */
 	while ((ch = GETOPT(argc, argv)) != NULL) {
 		GETOPT_SWITCH(ch) {
+		GETOPT_OPTARG("-c"):
+			if (opt_c != -1)
+				usage();
+
+			// read-in the chmod octal for the source socket
```
gperciva: Please use `/* comment */` instead of `//`.

libcperciva/util/sock.c
gperciva: Correction: perhaps something like "set the permission to ${sm} (if it is not -1)"

cperciva:
Can't you use `umask` for this purpose?  Using `chmod` after the socket is created can provide a misleading sense of security since another process might connect to the socket before the permissions are changed; but using umask will get the socket created with the right permissions in the first place.

wyattoday:
You're absolutely right. Also, using umask() has the added benefit of not having to modify the sock.c/sock.h files.

I'll rework the patch using umask().

gperciva:
Silly question: can a socket connect to a socket before `listen()` is called?

gperciva:
Be careful about the formatting of the man-page.  Right now, it displays as this:

```
       -c <source chmod>
              The access permissions of the source  socket  (if  it's  a  Unix
              socket).   For example: -c 777 Defaults to the system default of
              source sockets.
```

I think the "For example: -c 777" could be omitted.

wyattoday:
That's what I've been looking into, and it looks like no other process will be able to connect to the socket until the listen() call. Another process can, however, swoop in between the bind() and chmod() calls and listen() on the socket -- but then the listen() call in spiped will fail. So, I don't see it as a security problem.

Tell me if I'm missing something.

cperciva:
There's two possible problems here.  First, the semantics of SO_REUSEADDR on unix sockets are undefined; *as far as I know* all existing operating systems make it a no-op, so you're right that anyone else listening would stop us from listening.  But there's no guarantee that nobody will implement that flag in the future, or even that no obscure operating systems implement it already.

The second possible problem is a kernel race: Filesystem object ownership tends to be checked at a very different point in the kernel from socket properties.  I wouldn't be surprised if it's possible for a process to call `connect` before the `chmod` call happens, pass the permissions check, but not reach the "is anyone listening on this socket" check until after the `listen` call happens.  It would take some luck, of course; but concurrent systems are actively malicious, so I wouldn't want to assume that it could never happen.

cperciva:
It seems that even `umask` doesn't portably limit filesystem permissions on local domain sockets; since it's impossible to provide this functionality securely, I'm going to close this PR.

If anyone else is looking for this in the future, I recommend creating a directory with the appropriate permissions and having sockets created inside there.

PR #43, opened 2016-04-13
Makefiles bsd posix

gperciva:


spipe/Makefile.BSD
```diff
@@ -78,8 +78,6 @@ SRCS	+=	crypto_entropy.c
 SRCS	+=	crypto_verify_bytes.c
 IDIRS	+=	-I../libcperciva/crypto
 
-crypto_aes_aesni.o: crypto_aes_aesni.c cpusupport-config.h
```
gperciva: I must confess that I'm not 100% certain that removing this is safe -- it appears that this code wasn't used at all?  Was this perhaps only used when doing `make -F Makefile.BSD all`?

Removing it has no change on the makefiles produced via `make Makefiles`, and spiped still compiles, and (on my system) the `-maes` option is is correctly passed to `crypto_aes_aesni.o` (and only that file).

gperciva:
Updated PR to accommodate `cpusupport-config.h`.

gperciva:
huh, github claims that I commented on an outdated diff.  Reproduced here, about the last commit in the series:

> I must confess that I'm not 100% certain that removing this is safe -- it appears that this code wasn't  used at all?  Was this perhaps only used when doing `make -F Makefile.BSD all`?

> Removing it has no change on the makefiles produced via `make Makefiles`, and spiped still compiles, and (on my system) the `-maes` option is is correctly passed to `crypto_aes_aesni.o` (and only that file).

spipe/Makefile.BSD
cperciva: Yes, that compile rule is only used when running `make all` from Makefile.BSD.

cperciva:
Ok, looks generally plausible.  Three things I don't like:

1. Your sed is broken on FreeBSD.
2. I don't think we need line wrapping of the dependencies anyway.
3. I think the commits are easier to read if you leave "rebuild POSIX Makefiles" as a single commit at the end, rather than having all of the make targets change several times.

gperciva:
Updated PR.  With regard to https://github.com/Tarsnap/spiped/pull/43/commits/e0fa031e2a758ae99378ffe9438e9e75e9106623, I'm starting to understand the distain that some people have towards GNU extensions.

PR #44, opened 2016-04-15
spipe: make sure all the I/O has been completed.

ppentchev:
Hi,

Thanks for writing and maintaining spiped!

What do you think about this patch that makes spipe wait for the threads that it has spawned to actually finish their I/O before exiting after the remote socket has been closed?  As things are now, it is possible that the socket will be closed very soon after spipe has passed the data on its internal pipe to one of the pushbits threadsa, but the thread has not yet had a chance to output it.  For a more detailed explanation of an actual failed test case, take a look at http://devel.ringlet.net/security/spiped/pthread-race/

Thanks again for you're doing for spiped, tarsnap, FreeBSD, etc, and keep up the great work!

G'luck,
Peter

### DIFF
--- a/spipe/main.c
+++ b/spipe/main.c
@@ -1,6 +1,7 @@
 #include <sys/socket.h>
 
 #include <inttypes.h>
+#include <pthread.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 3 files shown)


cperciva:
Ooh, I wonder if this explains the occasional warnings I get from SSH when I close a connection?

cperciva:
I'm slightly concerned about this from the perspective of "what if we're being piped into a process which has stopped reading and our output buffer fills up".  Arguably the right solution in that case is to stall and not exit until we manage to finish writing the data, and I wouldn't mind if I was sure that we would only do that; but I'm not absolutely certain that we wouldn't leave a half-closed TCP connection...

cperciva:
I have some comments about the patch itself, but I'd like to be convinced of the behaviour we should be aiming for before worrying about the details of the code.

Hmm, why is this a problem in spipe but not spiped?

cperciva:
@ppentchev ping?

cperciva:
Closing this due to lack of feedback (and the patch currently conflicting).  Feel free to reopen -- I'm not opposed to this in principle, but I want to understand the changes enough to be confident that they're correct before I merge anything.

PR #45, opened 2016-04-15
Fix a couple of typographical errors.

ppentchev:
Hi,

Here's a trivial patch to fix a couple of typos.

Thanks in advance for your time!

G'luck,
Peter


cperciva:
Thanks!  This is $4 of bug bounties if you have a Tarsnap account.

PR #46, opened 2016-04-17
pass negative (sets default/max) listen() backlog

c0decafe:
tested on freebsd 10, but it seems to be common (best?) practice on linux too.

### DIFF
--- a/libcperciva/util/sock.c
+++ b/libcperciva/util/sock.c
@@ -327,7 +327,7 @@ sock_listener(const struct sock_addr * sa)
 	}
 
 	/* Mark the socket as listening. */
-	if (listen(s, 10)) {
+	if (listen(s, -1)) {
 		warnp("Error marking socket as listening");
 		goto err1;
 	}


cperciva:
POSIX says:

> If listen() is called with a backlog argument value that is less than 0, the function behaves as if it had been called with a backlog argument value of 0.

> A backlog argument of 0 may allow the socket to accept connections, in which case the length of the listen queue may be set to an implementation-defined minimum value.

So depending on the operating system, passing -1 here might make it not work at all.

If we're going to change it SOMAXCONN might make more sense.

c0decafe:
You are right and that was my initial approach.

Unfortunately that would be replacing one arbitrary value by another and one that cannot by tuned by sysctl(8) or equivalent.

We could of course use the likes of systctlbyname() to figure that out at runtime, but that will be messy to get right and portable.

I therefore opted for the -1

cperciva:
So, instead of a constant which might be wrong, you opted for something which the standard says *is* wrong?  Nope.

c0decafe:
why was 10 chosen, if you don't mind me asking?

cperciva:
It was pretty much arbitrary.  1 would have been too low; 100 seemed like it would be too large (especially since I'm using `select` and so can't handle huge numbers of connections at once).

c0decafe:
well, for me "implementation-defined minimum value" didn't sound too bad, especially since it doesn't apply to any of sane operating systems i know of.

Regardless, thanks for a lot for spiped, greatly simplified the deployment of a large db replication cluster.

cperciva:
> well, for me "implementation-defined minimum value" didn't sound too bad

You misunderstand; it *may* be set to an implementation-defined minimum value... or the socket might simply not be set to listen at all.  The former would be fine; the latter would be a problem.

Since my policy is "code to the standard" rather than "code to the implementations I have available", the fact that I don't know of any systems which handle this perversely is irrelevant.

PR #47, opened 2016-04-20
man: -f means disabling the Diffie-Hellman handshake

gperciva:
This arose from:
    http://mail.tarsnap.com/spiped/msg00002.html


gperciva:
Not an absolutely necessary piece of information, but I think it's worth being a bit more explicit about.

Issue #40, opened 2016-04-02
text in README.md about makefiles


cperciva:
```
Official releases on www.tarsnap.com/spiped.html have POSIX-compliant Makefiles; please use those.

:warning: The git repository (including the tag-based "release" snapshots on github) uses BSD Makefiles, which may or may not work on your operating system.
```

These comments have been out of date since `c46ceda5` I think?

gperciva:
Hmm.  Those comments were added in 29fe39b2 2015-10-21, whereas c46ceda5 was 2015-05-26.  We added this warning specifically in response to https://github.com/Tarsnap/spiped/issues/26 .

That said, the complaint was about spiped 1.5.0, which came before c46ceda5.  So perhaps we can delete the second paragraph (about the git repository), while keeping the first sentence (pointing people at the official releases).

cperciva:
Oh, I remember now.  Maybe we should just leave this alone until we release 1.6.0 (hmm, why haven't we done that yet?) and then remove it since people aren't likely to download *old* tarballs from github.

gperciva:
Sounds good to me.  I think 1.6.0 could wait until I've finished the test suite (including adding it to travis-ci).

PR #50, opened 2016-05-03
Readme redistribute

gperciva:


DESIGN.md
```diff
@@ -6,47 +6,63 @@ Encrypted protocol
 
 The client and server share a key file with 256 or more bits of entropy.  On
 launch, they read the key file and compute
+
     K = SHA256(key file).
 
 When a connection is established:
-C1. The client generates a 256-bit random value nonce_C and sends it.
```
gperciva: Not certain if you want to leave DESIGN in plain-text, or markdownify it.  I definitely think that README should be markdown (to display nicely on github), but there's an argument in favor of not changing DESIGN since it's only aimed at cryptographers (who will surely be familiar with plaintext math).  I can remove that change easily if you want.

I also noticed that BUILDING is plaintext; I'd be happy to markdownify that as well.  Also, BUILDING isn't linked (in terms of href) from README.md... basically, my rule of thumb was that if something is linked directly from README.md, it should look nice in the github interface, which means markdown or extremely simple plain text (which DESIGN and BUILDING do not qualify as).

gperciva:
Slightly updated `DESIGN.md`.  I forgot to mention earlier that this PR also addresses #40, and you can see it rendered at
https://github.com/Tarsnap/spiped/tree/readme-redistribute
https://github.com/Tarsnap/spiped/blob/readme-redistribute/DESIGN.md

README.md
```diff
@@ -1,26 +1,94 @@
-spiped
-------
+spiped and spipe
```
cperciva: I refer to the entire package as `spiped`, even though it now contains two utilities.  In the future there might be even more utilities, e.g., if asymmetric keying is added, but the package as a whole will still be "spiped".

gperciva:
Updated PR to keep the title as "spiped".

PR #51, opened 2016-05-08
Perform a graceful shutdown on SIGTERM

TimWolla:


spiped/dispatch.c
```diff
@@ -80,7 +82,7 @@ doaccept(struct accept_state * A)
 	int rc = 0;
 
 	/* If we can, accept a new connection. */
-	if ((A->nconn < A->nconn_max) && (A->accept_cookie == NULL)) {
+	if ((A->nconn < A->nconn_max) && (A->accept_cookie == NULL) && !(A->shutdown)) {
```
TimWolla: This does not work like I expect yet. This change should prevent accepting new connections once `SIGTERM` was received, can you tell me what I did wrong?

gperciva: `doaccept` seems to prepare for an incoming connection in advance, rather than choosing whether to accept a connection which is happening right now.  I would be tempted to call it "prepare_for_future_connection".

We can see this by adding a few `printf`s to the code:

```
$ ./spiped/spiped -d -s 127.0.0.1:8001 -t 127.0.0.2:8002 -k blank.txt -F
dispatch doaccept
    // wait a few seconds, then run "telnet 127.0.0.1 8001" in a different terminal
dispatch callback_gotconn
dispatch doaccept
```

In order to achieve what you want, I would start looking at `callback_gotconn` rather than `doaccept`.

That said, I'm not intimately familiar with the `spiped` code & design, and I'm not certain this is a good way to approach the problem.  I recommend waiting a few more days for @cperciva to chime in (this is a very busy time for him).

cperciva: > doaccept seems to prepare for an incoming connection in advance, rather than choosing whether to accept a connection which is happening right now. I would be tempted to call it "prepare_for_future_connection".

doaccept calls network_accept, which is a non-blocking version of accept.  The name is in analogy with the behaviour of accept, which waits until a connection arrives and then accepts it.

cperciva: @TimWolla I think what you're looking for here is network_accept_cancel.

TimWolla: @cperciva I think you are right :wink: . I just pushed a new commit that calls `network_accept_cancel` and a short tests confirms that spiped indeed does not accept a new connection.

TimWolla:
With the one issue fixed this pull request is ready for review. I can squash the commits if you want (either before or after review).

spiped/dispatch.c
```diff
@@ -23,6 +23,8 @@ struct accept_state {
 	int requirefps;
 	int nokeepalive;
 	int * conndone;
+	int only_one;
+	int shutdown;
```
cperciva: I'd call this `shutdown_requested`; the name `shutdown` is ambiguous (is it an adjective or a verb?).

spiped/dispatch.c
cperciva: Superfluous parentheses.

spiped/dispatch.c
```diff
@@ -174,7 +178,7 @@ void *
 dispatch_accept(int s, const char * tgt, double rtime, struct sock_addr ** sas,
```
cperciva: The comment here describing the function should be kept in sync with the .h file.

spiped/dispatch.h
```diff
@@ -19,17 +19,25 @@ struct sock_addr;
...
  * Stops and frees memory associated with the ${dispatch_cookie}.
  */
 void dispatch_shutdown(void *);
 
+/**
+ * request_shutdown(dispatch_cookie):
+ * Requests a shutdown: Stop accepting new connections and notify once
+ * every existing connection ended.
+ */
+void request_shutdown(void *);
```
cperciva: Should be named `dispatch_something`.

spiped/dispatch.c
```diff
@@ -125,6 +127,8 @@ callback_gotconn(void * cookie, int s)
 
 	/* We have gained a connection. */
 	A->nconn += 1;
+	if (A->only_one)
```
cperciva: This changes the behaviour of the `-1` option: Previously we would accept multiple connections and exit as soon as the any of them was closed; this makes spiped stop accepting connections once one has arrived.

@gperciva, that option was something you added for testing purposes, I think... does this affect your testing?  I don't mind changing the behaviour of the option given that it's documented as "for internal testing", but I don't want to break the tests...

spiped/dispatch.c
```diff
@@ -247,3 +253,23 @@ dispatch_shutdown(void * dispatch_cookie)
...
 	free(A);
 }
+
+/**
+ * request_shutdown(dispatch_cookie):
+ * Requests a shutdown: Stop accepting new connections and notify once
+ * every existing connection ended.
+ */
+void
+request_shutdown(void * dispatch_cookie)
```
cperciva: Needs comments explaining what we're doing.

spiped/main.c
```diff
@@ -15,6 +15,8 @@
 #include "dispatch.h"
 #include "proto_crypt.h"
 
+volatile int should_shutdown = 0;
```
cperciva: sig_atomic_t

spiped/main.c
```diff
@@ -225,15 +253,16 @@ main(int argc, char * argv[])
 		}
 	}
 
+	if (signal(SIGTERM, graceful_shutdown_handler) == SIG_ERR) {
+		warnp("Failed to bind SIGTERM signal handler");
```
cperciva: If this fails, we should exit.  This is different from when we bind signals in order to work around the docker bug; in that case we specifically *do not* want to exit since that's a workaround.  (Rule #1 of workarounds: Don't break anything else.)

spiped/main.c
```diff
@@ -292,12 +321,15 @@ main(int argc, char * argv[])
...
-	    sas_t, opt_d, opt_f, opt_g, opt_j, K, opt_n, opt_o,
-	    opt_1 ? &conndone : NULL)) == NULL) {
+	    sas_t, opt_d, opt_f, opt_g, opt_j, K, opt_n, opt_o, opt_1,
+	    &conndone)) == NULL) {
 		warnp("Failed to initialize connection acceptor");
 		goto err5;
 	}
 
+	/* Check periodically whether a signal was received. */
+	events_timer_register_double(graceful_shutdown, dispatch_cookie, 1.0);
```
cperciva: @gperciva might want to record the cookie so that he can deregister the callback in the event of a clean exit so that it isn't flagged as a memory leak.

cperciva:
Please address the comments above (with feedback from @gperciva where relevant), then squash and I'll take another look.

spiped/dispatch.c
gperciva: The tests aren't in spiped yet, and it's easy enough for me to delete the one that tests opening two connections and closing one.  I think the main reason you wanted me to add that test was so that we could check if memory from active connections was being de-allocated before quitting.

gperciva: On second thought, it *is* nice being able to test that cleanup, so I'd rather leave `-1` unchanged.

spiped/dispatch.c
TimWolla: You are talking about the function body, right?
Am 14.05.2016 23:31 schrieb "Colin Percival" <notifications@github.com>:

spiped/main.c
TimWolla: This is not something I should change, but @gperciva might do after the
merge, right?
Am 14.05.2016 23:34 schrieb "Colin Percival" <notifications@github.com>:

spiped/main.c
TimWolla: Is it `volatile sig_atomic_t` or just `sig_atomic_t`?
Am 14.05.2016 23:31 schrieb "Colin Percival" <notifications@github.com>:

spiped/dispatch.c
cperciva: Yes, comments inside the body of the function.

spiped/main.c
cperciva: Yes, it should be `volatile sig_atomic_t`.

spiped/main.c
cperciva: Correct.  I was just drawing it to his attention in case he wanted to say "oh no, this will break everything I have planned". :-)

gperciva: Nah, won't break stuff.

spiped/dispatch.c
```diff
@@ -247,3 +253,23 @@ dispatch_shutdown(void * dispatch_cookie)
...
+{
+	struct accept_state * A = dispatch_cookie;
+
+	A->shutdown = 1;
+
+	if (A->accept_cookie != NULL)
+		network_accept_cancel(A->accept_cookie);
+		A->accept_cookie = NULL;
```
TimWolla: I just noticed that you missed the missing braces at this place :smile:

TimWolla:
Updated.

TimWolla:
Just force pushed to fix a typo in a comment.

spiped/dispatch.c
```diff
@@ -166,15 +172,16 @@ callback_gotconn(void * cookie, int s)
  * forward secrecy.  If ${requirefps} is non-zero, require that both ends use
  * perfect forward secrecy.  Enable transport layer keep-alives (if applicable)
  * if and only if ${nokeepalive} is zero.  Drop connections if the handshake or
- * connecting to the target takes more than ${timeo} seconds.  If ${conndone}
- * is not NULL, set it to non-zero value when a connection closes.  Returns a
- * cookie which can be passed to dispatch_shutdown.
+ * connecting to the target takes more than ${timeo} seconds.  If ${only_one}
+ * is non-zero request a shutdown once the first connection is acccepted.  If
```
cperciva: "If ${only_one} is non-zero request a shutdown once the first connection is acccepted."

Leaving aside the typo (acccepted), I don't think this is accurate?

TimWolla: @cperciva You're right. May I suggest:

> … If ${only_one} is a non-zero value ${conndone} is set to a non-zero value once a connection closes.  If a shutdown was requested and every connection ended ${conndone} is set to a non-zero value as well. …

gperciva: I suggest:
"If ${only_one} is a non-zero**, then** ${conndone} is..."
"... every connection ended**, then** ${conndone}..."
The original grammar is not strictly speaking incorrect, but the comma and "then" makes it easier to skim the sentence.

cperciva: I think this might be clearer:

> If ${only_one} is a non-zero value, then ${conndone} is set to a non-zero value as soon as a connection closes.  If dispatch_request_shutdown is called then ${conndone} is set to a non-zero value as soon as there are no active connections.

spiped/dispatch.c
```diff
@@ -166,15 +172,16 @@ callback_gotconn(void * cookie, int s)
...
+ * a shutdown was requested and every connection ended ${conndone} is set to a
+ * non-zero value.  Returns a cookie which can be passed to dispatch_shutdown.
```
cperciva: ... "and dispatch_request_shutdown".

TimWolla:
Updated.

cperciva:
@gperciva @TimWolla Can one of you also update CHANGELOG and spiped/spiped.1 to reflect the new SIGTERM handling?

PR #53, opened 2016-05-17
Proto conn cookie

gperciva:


gperciva:
This fixes a memory leak when opening 2 connections (with `spiped -1`) and closing one of them.

There are some merge conflicts #51, but I can fix that once #51 is accepted.  I recommend doing that one first.  :)

spiped/dispatch.c
```diff
@@ -132,11 +163,34 @@ callback_gotconn(void * cookie, int s)
...
-		goto err2;
+		goto err3;
+	}
+
+	/* Add conn_list_node node_new to accept_state A. */
+	if (A->conn_cookies == NULL) {
+		/* Nothing in list yet. */
+		A->conn_cookies = node_new;
+	} else {
+		/* Find end of list. */
```
TimWolla: You could always insert in the first place to avoid the loop, no?

gperciva: Good idea!  Also, that way we don't need to have any special case for the list not existing yet.

spiped/dispatch.c
```diff
@@ -130,20 +163,37 @@ callback_gotconn(void * cookie, int s)
...
+	    A->nofps, A->requirefps, A->nokeepalive, A->K, A->timeo,
+	    callback_conndied, A)) == NULL) {
 		warnp("Failure setting up new connection");
-		goto err2;
+		goto err3;
 	}
 
+	/* Link node_new to the beginning of the conn_cookies list. */
+	if (A->conn_cookies != NULL)
+		A->conn_cookies = node_new;
```
TimWolla: This looks incorrect to me. If `A->conn_cookies` previously was `NULL` it would link `node_new` to itself.

gperciva:
Updated PR, thanks for the warning about the missing `->prev`!

gperciva:
Updated to apply cleanly to master (i.e. after #51 was merged).

proto/proto_conn.h
```diff
@@ -17,9 +17,16 @@ struct sock_addr;
...
- * once it is no longer needed.
+ * once it is no longer needed.  Return a cookie which can be passed to
+ * proto_conn_drop.
  */
-int proto_conn_create(int, struct sock_addr **, int, int, int, int,
-    const struct proto_secret *, double, int (*)(void *), void *);
+void * proto_conn_create(int, struct sock_addr **, int, int, int, int,
+    const struct proto_secret *, double, int (*)(void *, void *), void *);
+
+/** proto_conn_drop(conn_cookie):
```
cperciva: formatting

spiped/dispatch.c
```diff
@@ -94,13 +103,34 @@ doaccept(struct accept_state * A)
...
+	struct conn_list_node * node_ptr;
+
+	assert(A->conn_cookies != NULL);
 
 	/* We've lost a connection. */
 	A->nconn -= 1;
 
+	/* Remove conn_cookie from the list of conn_cookies. */
+	node_ptr = A->conn_cookies;
+	while (node_ptr->conn_cookie != conn_cookie)
```
cperciva: Why are you walking the list?  Why not pass the node pointer as the cookie to be returned to us?

proto/proto_conn.h
gperciva: wow, it took me 2 whole minutes to spot the problem, even after you'd isolated the line!  (I actually thought you were complaining about `proto_conn_create`, but couldn't spot any problems there, either)

gperciva:
Updated PR temporarily (messy history).  @cperciva wrote (on an outdated diff):
> Why are you walking the list? Why not pass the node pointer as the cookie to be returned to us?

I didn't understand the concern, but my code comments were a bit unclear.  Hopefully this next version will clarify what (I think) I'm doing.

spiped/dispatch.c
```diff
@@ -94,13 +103,36 @@ doaccept(struct accept_state * A)
...
 {
 	struct accept_state * A = cookie;
+	struct conn_list_node * node_ptr;
+
+	assert(A->conn_cookies != NULL);
 
 	/* We've lost a connection. */
 	A->nconn -= 1;
 
+	/* Find the node that contains conn_cookie. */
```
cperciva: My question still remains.  You're walking a linked list looking for something here; walking linked lists looking for something is slow.

spiped/dispatch.c
```diff
@@ -136,20 +169,37 @@ callback_gotconn(void * cookie, int s)
...
+	/* Create new conn_list_node. */
+	if ((node_new = malloc(sizeof(struct conn_list_node))) == NULL)
+		goto err2;
+	node_new->prev = NULL;
+	node_new->next = NULL;
+
 	/* Create a new connection. */
-	if (proto_conn_create(s, sas, A->decr, A->nofps, A->requirefps,
-	    A->nokeepalive, A->K, A->timeo, callback_conndied, A)) {
+	if ((node_new->conn_cookie = proto_conn_create(s, sas, A->decr,
```
cperciva: If you put a pointer to A inside struct conn_list_node and then passed node_new as the cookie to be passed back to callback_conndied, you wouldn't need to search the linked list.

gperciva:
Much nicer solution!  Updated PR; one of the previous commits in this batch has changed (now we no longer return two pointers in the "dead" callback).  History of this PR is good, so if it's acceptable then it can be merged.

spiped/dispatch.c
```diff
@@ -95,13 +105,31 @@ doaccept(struct accept_state * A)
 
 /* A connection has closed.  Accept more if necessary. */
 static int
-callback_conndied(void * cookie)
+callback_conndied(void * cookie_node)
```
cperciva: s/cookie_node/cookie/

No need to change it since there's only one cookie.

spiped/dispatch.c
```diff
@@ -95,13 +105,31 @@ doaccept(struct accept_state * A)
...
+
+	assert(A->conn_cookies != NULL);
 
 	/* We've lost a connection. */
 	A->nconn -= 1;
 
+	/* Remove the closed connection from the list of conn_cookies. */
+	if (node_ptr == A->conn_cookies) {
+		/* Closed conn_cookie is first in the list. */
+		A->conn_cookies = node_ptr->next;
```
cperciva: What about `node_ptr->next->prev` ?

spiped/dispatch.c
```diff
@@ -95,13 +105,31 @@ doaccept(struct accept_state * A)
...
 /* A connection has closed.  Accept more if necessary. */
 static int
-callback_conndied(void * cookie)
+callback_conndied(void * cookie_node)
 {
-	struct accept_state * A = cookie;
+	struct conn_list_node * node_ptr = cookie_node;
+	struct accept_state * A = node_ptr->A;
+
+	assert(A->conn_cookies != NULL);
```
cperciva: Needs a comment to explain why.

spiped/dispatch.c
```diff
@@ -137,20 +166,38 @@ callback_gotconn(void * cookie, int s)
...
+	    A->nofps, A->requirefps, A->nokeepalive, A->K, A->timeo,
+	    callback_conndied, node_new)) == NULL) {
 		warnp("Failure setting up new connection");
-		goto err2;
+		goto err3;
 	}
 
+	/* Link node_new to the beginning of the conn_cookies list. */
+	if (A->conn_cookies != NULL)
+		A->conn_cookies->prev = node_new;
```
cperciva: Since you're initializing `node_new->next = NULL` earlier, you could put both of these lines under the `if`.

Also, I like spelling it `node_new->next->prev = node_new`.

cperciva:
Aside from the above, can you change the dates so that the commits are in order?  It's confusing seeing things change in random order.

spiped/dispatch.c
gperciva: Yeah, but it's a different cookie than all the other `callback_*()` functions in that `.c` file...?   Principle of least surprise?

*shrug*
ok, I'll un-change it.

cperciva: We call it the principle of least astonishment (POLA).  But to address the question, that's why the first local variable declaration in cookie-using functions should be `struct foo * bar = cookie;`.

... I guess that could do with documenting in STYLE, actually.

gperciva:
Updated PR, splitting the "track active connections" into two commits because I really cannot be trusted to write list-handling code.  Makes sense to make it as minimal as possible.
(well, close to "minimal".  I could have put the `dispatch_shutdown` material in a separate commit, but I thought that would be overkill)

spiped/dispatch.c
```diff
@@ -250,6 +302,16 @@ void
...
 	struct accept_state * A = dispatch_cookie;
+	struct conn_list_node * ptr;
+	struct conn_list_node * ptr_next;
+
+	/* Shutdown any open connections. */
+	ptr = A->conn_cookies;
+	while (ptr != NULL) {
+		ptr_next = ptr->next;
+		proto_conn_drop(ptr->conn_cookie);
+		ptr = ptr_next;
```
cperciva: Shouldn't you be freeing the linked list nodes here?

gperciva: `proto_conn_drop` calls back to `callback_conndied()`, which frees the linked list nodes.  I could add a comment to that effect?  Or would you rather have a different code design?

cperciva: Oh... hmm.  Yeah, we definitely need a comment to that effect.  Or possibly change `proto_conn_drop` to something like `proto_conn_free` which cleans up but doesn't perform the callback -- I've learned that it's possible to tie oneself in knots very easily by having "cancel"/"shutdown" functions make callbacks -- but in this case I don't think that's necessary.  (The scenario we absolutely must avoid is having proto_conn_drop call back to callback_conndied and then callback_conndied makes more calls to proto_conn_drop, since that could give us stack-overflowing recursion; but I'm 99% certain that we're not doing anything like that here.)

Also, I think that means this code can be simplified, to something along the lines of
`while (A->conn_cookies != NULL) proto_conn_drop(A->conn_cookies->conn_cookie);` ?

gperciva:
Updated PR.  I separated the "tracking" and "dropping any remaining connections" functionality, as well as cleaning up and explaining the "dropping" part better.

I initially did add a `proto_conn_free`, but since it was almost exactly the same thing as `proto_conn_drop`, that felt like a "risky" design pattern -- what if somebody only modified one of those files?  I was quite happy when I realized that I could use `proto_conn_drop` and `callback_conndied` and avoid having almost-identical functions.

PR #54, opened 2016-05-19
Document the SIGTERM behaviour

TimWolla:
see 8f22fea0485ac09d031d9e6c4b05b73690683725

### DIFF
--- a/spiped/spiped.1
+++ b/spiped/spiped.1
@@ -153,5 +153,15 @@ Print version number.
 .TP
 .B \-1
 Exit after a connection closes.  Used for internal testing.
+.SH SIGNALS
+spiped provides special treatment of the following signals:
+.TP
+.B SIGTERM
+On receipt of the
+.I SIGTERM
+signal
+.B spiped
+will stop accepting new connections and exit once there are
+no active connections left.
 .SH SEE ALSO
 .BR spipe (1).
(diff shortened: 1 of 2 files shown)


spiped/spiped.1
```diff
@@ -153,5 +153,13 @@ Print version number.
 .TP
 .B \-1
 Exit after a connection closes.  Used for internal testing.
+.SH SIGNALS
+.B SIGTERM
```
gperciva: hmm, for what it's worth, tarsnap.1 uses these commands:

```
.Sh SIGNALS
.Nm
provides special treatment of the following signals:
.Bl -tag -width "SIGUSR2"
.It SIGUSR1 & SIGINFO
On receipt of the SIGUSR1 signal or (on platforms where it exists) the
SIGINFO signal,
```

which results in it being formatted as:

```
SIGNALS
     tarsnap provides special treatment of the following signals:

     SIGUSR1 & SIGINFO
              On receipt of the SIGUSR1 signal or (on platforms where it
              exists) the SIGINFO signal,
```

I personally like that formatting a bit more than the proposed change, which doesn't indent the explanation paragraph relative to the `SIGTERM`:

```
SIGNALS
       SIGTERM When spiped receives the SIGTERM signal it will stop  accepting
       new connections and exit once there are no active connections left.
```

TimWolla: @gperciva Go ahead please, I don't want to spend time refining this pull request at the moment. I agree that my current formatting is pretty ugly, though. I tried to find man pages on my system that have a `SIGNALS` section, but could not find any.

TimWolla: @gperciva Oh, nevermind. I found the formatting mistake, force pushing any second.

PR #55, opened 2016-05-19
spiped: track & clean up graceful shutdown cookie

gperciva:
events_timer_cancel(graceful_shutdown_timer_cookie);
must happen before

    dispatch_shutdown(dispatch_cookie);

because otherwise the timer could trigger after the shutdown (which would try
to use dispatch_cookie after it was freed).

### DIFF
--- a/spiped/main.c
+++ b/spiped/main.c
@@ -16,6 +16,7 @@
 #include "proto_crypt.h"
 
 volatile sig_atomic_t should_shutdown = 0;
+static void * graceful_shutdown_timer_cookie = NULL;
 
 static void
 usage(void)
@@ -64,7 +65,8 @@ graceful_shutdown(void * cookie)
 	if (should_shutdown)
 		dispatch_request_shutdown(A);
 	else
-		events_timer_register_double(graceful_shutdown, A, 1.0);
+		graceful_shutdown_timer_cookie = events_timer_register_double(
+		    graceful_shutdown, A, 1.0);
 
 	return 0;
 }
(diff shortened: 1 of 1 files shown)


gperciva:
I don't like using a global variable for this, but I thought it might be overkill to replace

    graceful_shutdown(void * cookie)
    {

with an uber-cookie which contained both `A` and `graceful_shutdown_timer_cookie`.  If you'd rather have that solution, though, I can whip it up quickly.

gperciva:
Also, should I mention the relative order of `events_timer_cancel` and `dispatch_shutdown` in a comment in `main.c`?

spiped/main.c
```diff
@@ -64,7 +65,8 @@ graceful_shutdown(void * cookie)
 	if (should_shutdown)
 		dispatch_request_shutdown(A);
```
cperciva: If we get a signal and call into dispatch_request_shutdown here...

spiped/main.c
```diff
@@ -340,6 +343,9 @@ main(int argc, char * argv[])
 		goto err6;
 	}
 
+	/* Deregister the graceful_shutdown timer. */
+	events_timer_cancel(graceful_shutdown_timer_cookie);
```
cperciva: ... then don't we end up trying to cancel a timer which doesn't exist here?

I suggest setting to NULL when the callback is invoked and then checking for NULL here.

gperciva:
PR updated with messy history.

gperciva:
updated PR to fix embarrassing mistake

PR #58, opened 2016-06-03
Style comment return

gperciva:


gperciva:
This isn't explicitly discussed in STYLE, but before this commit,

```
$ git grep "Return" | wc -l
81
$ git grep "Returns" | wc -l
5
```
so I figure you'd prefer it without an "s".

PR #61, opened 2016-06-22
Build: use -std=c99 for sha256.c

gperciva:


gperciva:
I considered adding this as `CFLAGS_DEFAULT`, but since you went with a targeted approach in https://github.com/Tarsnap/libcperciva/commit/53d158a17ed9e0dcf073414838d0a16234ca9485 , I figured I should do the same thing here.

gperciva:
BTW, without this commit,

    make CC=c99

works just fine.  Even on gcc-4.8.4 (default in Ubuntu 14.04).

gperciva:
On second thought, http://pubs.opengroup.org/onlinepubs/9699919799/utilities/c99.html doesn't support the `-std=` option.  So maybe this isn't the best solution.  (in which case, we might want to revisit its use in the libcperciva test suite)

Y'know, it might be useful if there was some way to **auto**matically detect some **conf**iguration details of the current OS.  Like, such a **auto**matic **tool** could **s**implify the task of figuring out what&when to add stuff to CFLAGS.  That way, we could put platform-specific stuff in some separate top-level files, and have a simpler Makefile logic.  Oh well, maybe somebody will build a time machine and go back to 1994 to start working a solution for this problem.  :)

cperciva:
I think this should be part of CFLAGS_POSIX.  As you say, autodetection. :-)

gperciva:
I'd call this "manual detection", not "autodetection".  :P

cperciva:
I wrote the code for this myself.

PR #62, opened 2016-06-23
Print error message if given an illegal option

gperciva:


gperciva:
Ok, so glibc and libc's getopt have different "I don't understand that argument" messages.  And our magic gitopt doesn't print anything by default, so we need to print an error, otherwise 1) users might be confused, and more importantly 2) there wouldn't be anything to grep for, if we need to check the existence of other options in the future.

Given `spiped -z` and `spiped --foo` (in version 1.5), glibc prints:

    spiped: invalid option -- 'z'
    spiped: invalid option -- '-'

while FreeBSD's libc prints:

    spiped: illegal option -- z
    spiped: illegal option -- -

To be Canadian, I gone with:

    spipe: illegal option -- -z
    spipe: illegal option -- --foo

Yes, I could print out `ch + 1` to avoid the leading `-`, but I actually think it's easier to read if we include that -.  Especially if somebody has lots of options in there, and just typo'd the `-k` or something like that.

spipe/main.c
```diff
@@ -108,6 +108,7 @@ main(int argc, char * argv[])
 			warn0("Missing argument to %s\n", ch);
 			/* FALLTHROUGH */
```
gperciva: ah, I'd forgotten about that FALLTHROUGH.  I'll replace that with `usage()`.

gperciva:
Updated PR

gperciva:
I've made a note to update the other magic getopt projects if/when this PR is accepted.

PR #64, opened 2016-06-28
Fix compile

gperciva:


Makefile
```diff
@@ -6,8 +6,8 @@ CFLAGS_DEFAULT=	-O2
 
 all: cpusupport-config.h
 	export CFLAGS="$${CFLAGS:-${CFLAGS_DEFAULT}}";	\
-	export LDADD_POSIX=`export CC="${CC}"; cd libcperciva/POSIX && command -p sh posix-l.sh`;	\
-	export CFLAGS_POSIX=`export CC="${CC}"; cd libcperciva/POSIX && command -p sh posix-cflags.sh`;	\
+	export LDADD_POSIX=`export CC=\`which "${CC}"\`; cd libcperciva/POSIX && command -p sh posix-l.sh`;	\
```
cperciva: This can't really be necessary, can it?

libcperciva/POSIX/posix-cflags.sh
```diff
@@ -1,18 +1,17 @@
...
 	echo "WARNING: POSIX violation: <sys/socket.h> not defining MSG_NOSIGNAL" 1>&2
 fi
-if ! ${CC} -D_POSIX_C_SOURCE=200809L posix-clock_realtime.c 2>/dev/null; then
+if ! "${CC}" -D_POSIX_C_SOURCE=200809L posix-clock_realtime.c 2>/dev/null; then
 	[ ${FIRST} = "NO" ] && printf " "; FIRST=NO
 	printf %s "-DPOSIXFAIL_CLOCK_REALTIME"
 	echo "WARNING: POSIX violation: <time.h> not defining CLOCK_REALTIME" 1>&2
 fi
-if ! ${CC} -D_POSIX_C_SOURCE=200809L posix-restrict.c 2>/dev/null; then
-	echo "WARNING: POSIX violation: ${CC} does not accept 'restrict' keyword" 1>&2
```
cperciva: Where did this warning message go?

libcperciva/POSIX/posix-cflags.sh
```diff
@@ -14,6 +14,8 @@ if ! "${CC}" -D_POSIX_C_SOURCE=200809L posix-restrict.c 2>/dev/null; then
 	if "${CC}" -D_POSIX_C_SOURCE=200809L -std=c99 posix-restrict.c ; then
 		[ ${FIRST} = "NO" ] && printf " "; FIRST=NO
 		printf %s "-std=c99"
+	else
+		echo "ERROR: adding -std=c99 does not fix 'restrict'; cannot continue". 1>&2
```
cperciva: A bit misleading, since we will in fact continue (until the compiler chokes on sha256.c)...

Makefile
gperciva: Passing it to `posix-l.sh`?  No.  Passing it to `posix-cflags.sh`?  Yes.  And since I did it for that, I figured that I might as well do it to the previous line.

Another option would be to pass in $PATH, but I figured that adjusting the ${CC} variable was a "smaller" change.  No?

gperciva:
I've made a small (not quite minimal, since the `which` for `posix-l.sh` is debateable) PR that only does the compile-fix: #65.  I think the double quotes and error message would be nice, but that can wait until master is compile-able again (and until I've fixed the actual error).

PR #65, opened 2016-06-28
build: pass full path of $CC to posix scripts

gperciva:


gperciva:
In more detail, on travis-ci OSX,

```
WARNING: POSIX violation: make's CC doesn't understand -lrt
WARNING: POSIX violation: make's CC doesn't understand -lxnet
----- CC:
/usr/local/bin/gcc-4.9
/usr/bin:/bin:/usr/sbin:/sbin
-----
WARNING: POSIX violation: <sys/socket.h> not defining MSG_NOSIGNAL
```
(err, the second line is $PATH.  Not going to fix that, since it would take another 60 minutes for it to actually run that code.  But you can check the git branch if you want confirmation.)

https://github.com/Tarsnap/spiped/tree/test-travis
https://travis-ci.org/Tarsnap/spiped/builds/140897448

cperciva:
So, looks like this is a bug in the shell which `make` is invoking.  The following should print the same line twice:
```
echo $PATH; command -p sh -c 'echo $PATH'
```
but on OS X `command -p` is wiping the $PATH -- hence it being able to find gcc-4.9 from the makefile but not from the shell script.

I'm guessing this is bash?

gperciva:
Yes, travis-CI uses OSX 10.9.5 by default, which (according to a random webpage I found) uses bash by default.

I can confirm that GNU bash 4.3.11 on Ubuntu prints different stuff:
```
td@gin: /home/gperciva
$ echo $PATH
/home/td/node_modules/.bin/:/home/td/.local/bin:/home/td/.local/clang-3.8/bin:/home/td/.bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
td@gin: /home/gperciva
$ command -p sh -c 'echo $PATH'
/bin:/usr/bin
```

PR #66, opened 2016-06-29
Spipe error path

gperciva:


spipe/main.c
```diff
@@ -18,14 +18,20 @@
 
 #include "pushbits.h"
 
+struct accept_state {
```
cperciva: Why create a structure when we could simply pass a pointer to the int itself?

gperciva: 1) I was trying to keep this as similar to https://github.com/Tarsnap/spiped/commit/49470ac389c187a69311fe024efa9f549b5cfe9b as possible.

2) I figured that you'd rather see

```
static int
callback_conndied(void * cookie)
{
```
instead of
```
static int
callback_conndied(void * cookie)
{
        struct int * conndone = cookie;
```

3) although spiped has an intermediate layer between `int main` and the `callback_conndone` (namely `dispatch.c`), spipe doesn't.  So if there's going to be any `void * cookie`, we need to prepare that inside `int main`.

cperciva: I think the spiped and spipe code paths are sufficiently different here that trying to make the patches look similar is pointless.  In spiped we store the pointer to conndone in the accept_state structure because it needs to be passed to every connection we establish; but in spipe we're creating a connection from main() and we can just provide the pointer right there without stashing it in a structure temporarily.

And casting an (int *) to a (void *) is just fine, so nothing to worry about there.

gperciva:
Ok, done.  Keeping the history truthful was a bit confusing, since the 2nd commit added a bunch of lines that the 3rd promptly deleted, so I squashed tonight's edits into the 2nd commit.

spipe/main.c
```diff
@@ -176,16 +179,23 @@ main(int argc, char * argv[])
...
 	/* Loop until we die. */
-	do {
-		if (events_run()) {
-			warnp("Error running event loop");
-			exit(1);
-		}
-	} while (1);
+	if (events_spin(&conndone)) {
+		warnp("Error running event loop");
+		exit(1);
```
cperciva: Did you want this to be a `goto errN` as well?  Or was the idea that if we die at this point there's no way we can possibly clean everything up, so don't worry about keeping valgrind happy?

gperciva: I do want it to be `goto errN` eventually, but I haven't got a good way of cleaning up after `pushbits`.  I'm implementing the error path in stages (as I did with `spiped`), so I'd prefer to get this merged, then I'll take another look at my various branches that attacked `pushbits` and discuss it with you tomorrow morning.

PR #67, opened 2016-07-01
Coverity: switch to new format in .travis.yml

gperciva:


gperciva:
Judging from
- this commit on the `coverity-scan` branch: https://travis-ci.org/Tarsnap/spiped/builds/141498698   
- this commit on the `coverity_scan-proposed` branch: https://travis-ci.org/Tarsnap/spiped/builds/141498865

this does what we want.  (also, thanks to the tip in https://github.com/travis-ci/travis-ci/issues/1975#issuecomment-216604414 , which came after @cperciva's tip in that issue)

Namely,
- if the branch is coverity_scan and the job number is not 1, exit
- otherwise, do the tests as normal.  (note that we check the job number *before* updating apt-get/brew and installing software)
- also, if the branch is coverity_scan, don't do a normal `make && make test`.

cperciva:
This may be a dumb question, but why are the encoded variables changing?

gperciva:
Do you mean the `env: global: - secure` string?  It looks like http://scan.coverity.com generates a new one every time you visit the page.  (at the time, I just assumed that they'd changed their hash for some reason, so adjusted the value in our `.travis.yml`)

> Note: COVERITY_SCAN_TOKEN is encrypted and encoded in the secure: declaration using your private Project token \*\*\*hunter2*** and may be different each time this page is loaded. See Travis CI's documentation on secure environment variables or review the Travis library for more information.
https://scan.coverity.com/projects/tarsnap-spiped/builds/new?tab=travis_ci

cperciva:
Ah, I see.  Since this is just a case of "Coverity likes generating new blobs for no reason", let's leave it unchanged; no point confusing the git history.

gperciva:
Updated PR, keeping the old `secure` blob, and the test looks ok.

gperciva:
err, I'm confused.  Why do you want it called `coverity-scan` instead of `coverity_scan` ?  I figured that matching the docs (and therefore any stackoverflow / etc. posts) would eliminate one possible source of confusion.

cperciva:
No particular reason, except that I had a branch named `coverity-scan` which I was using for this purpose already, and my "merge everything to the Coverity branch" script was coded to use that name.  I didn't think there was any reason to change the branch name we were already using.

PR #68, opened 2016-07-02
Include headers

gperciva:


proto/proto_crypt.h
```diff
@@ -1,6 +1,7 @@
 #ifndef _PCRYPT_H_
 #define _PCRYPT_H_
 
+#include <stddef.h>
```
cperciva: Not needed since we're getting size_t from unistd.h, I think?

proto/proto_handshake.c
```diff
@@ -1,3 +1,5 @@
+#include <sys/types.h>
```
cperciva: - <sys/types.h>: ssize_t

STYLE says get this from unistd.h.

gperciva:
Updated PR as you suggested.  Since it was only 2 small changes, I rebased the history instead of leaving them as separate patches.

PR #69, opened 2016-08-06
Libcperciva update

gperciva:


gperciva:
These all came from https://github.com/Tarsnap/libcperciva/pull/27

PR #74, opened 2016-08-16
spipe: split pushbits() || pushbits() into separate commands

gperciva:


gperciva:
This is mainly a test of the "protected branch" functionality.  And I do indeed see that, as I type this, I cannot click on "merge pull request" because the tests are still running.

gperciva:
also, I was rejected from pushing directly to master, so it looks like it's working.

gperciva:
err, clarification: I tried pushing a 1-line change to the README to master.  I wouldn't try to push a code change to master.  :)

PR #75, opened 2016-08-16
Man1dir install

gperciva:


cperciva:
This looks good, but apparently I can't merge it.  Is there a conflict?

gperciva:
hmm, apparently that's a "feature" of the "protected branch" stuff.  (or rather, of the "require status checks" portion of protected branches)

Argument is this: suppose there's two PRs, A and B.  Each one individually passes the tests, but if they're both applied, they will break something.  Since the goal of "protected branches + status checks" is that `master` should never be broken, it makes sense that you can't accept both PRs unless the second one is updated to include the first one.

This could get annoying, so I'll remove the "status checks" portion.  This should mean that I still can't push directly to master, but in theory you could merge a PR which was still in the process of being tested.

gperciva:
nope, never mind.  There's a separate checkbox for "require branch to be up-to-date", which I've unchecked, so presumably we can theoretically break stuff with a malicious A+B PR set, but it won't get in the way of our normal way of working.

cperciva:
Thanks.  I'm a bit confused as to why github decided it needed to re-run the tests now, but I guess I can wait...

gperciva:
Sorry, that was me messing around.  I saw the "update from master" button, clicked that, and then noticed that the history had a bunch of extra "merge commits".  So I rebased my branch locally, then uploaded that, so that the git history would look normal.  That triggered the usual re-test.

PR #76, opened 2016-08-16
Fix warnings

gperciva:


spiped/main.c
```diff
@@ -19,6 +19,9 @@
 volatile sig_atomic_t should_shutdown = 0;
 static void * graceful_shutdown_timer_cookie = NULL;
 
+/* Function prototype. */
+int graceful_shutdown(void * cookie);
```
cperciva: Why isn't `graceful_shutdown` declared as static?

cperciva:
Apparently I left a comment on the commit rather than the PR.  Why isn't `graceful_shutdown` static?

Also, while I'm here, a style bug I didn't notice earlier: `return 0` -> `return (0)`.

spiped/main.c
gperciva: Probably because there was some doubt as to whether you can give a pointer to a static function to another C file.  At least, that's why I assumed it wasn't `static` from the beginning -- it seems counter-intuitive to say "this function is only going to be used in main.c"[*] and then pass it to `events_timer_register_double()`.

[*] yes, I know that's not **actually** what `static` means; the actual definition has to do with the linker and translation units.

Casual googling suggests that it's safe to do that, and testing on ubuntu and freebsd looks.

gperciva:
Updated PR.

Comments on a specific commit are still visible in the PR "conversation".  I put the style into #77.

PR #79, opened 2016-08-27
Fix warnings

gperciva:


gperciva:
This is the last set of warnings I have queued up for spiped.

gperciva:
BTW, if you can think of any non-ridiculous titles for these commits, I'm all for changing them -- and I've got a lot more "explicitify these implicit casts" commits coming for the tarsnap client, so this could safe you time later on.

cperciva:
The titles look fine to me.  And BTW the commit messages are great -- explaining why each cast is safe is exactly what we need in case we look back at this many years later.

gperciva:
The commit messages are partly for the history, but mostly because I got fed up with missing some corner cases and making the wrong fixes.  If I can write a "proof" that the fix is good, then it's less likely that I've screwed up.  :)

PR #81, opened 2016-09-02
Fix warnings

gperciva:


gperciva:
These are the last warnings for spiped (with the standard clang `-Weverything -Wno-` options) in clang 3.9.0.

PR #86, opened 2016-09-20
Use apt addon in .travis.yml

TimWolla:
This might or might not simplify the travis configuration in your opinion. In my opinion it does.


cperciva:
LGTM... @gperciva, I think you wrote those lines?  Any particular reason you wrote them the way you did (aside from possibly "didn't know about the apt addon")?

gperciva:
The travis-ci docs don't use the `apt` method in the first example in https://docs.travis-ci.com/user/installing-dependencies/ , and when that page mentions the "APT addon", they also say
> If your requirements goes beyond the normal installation, please use another method described above.

All in all, it looked like the `before_install:` option was more widely supported and tested.  I also wasn't certain how this played with the osx builds.  It looks like osx silently ignores the apt lines, which is what we want although I was hoping to find that actually documented somewhere.
(hmm, it looks like https://docs.travis-ci.com/user/multi-os/index.html#example-multi-os-build-matrix does exactly what is proposed)

I have no objection to this change; I was going to propose a similar thing if `apt` seemed more officially supported, and/or they added a `brew` option.

PR #87, opened 2016-09-20
Switch to container based Travis infrastructure

TimWolla:
This is a superset of #86, which is a necessary requirement for this one. 

Using the container based infrastructure allows for faster builds (not that it would really matter for this project) and reduces the load on the free service provided by Travis and thus should be preferred.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: c
-sudo: required
-dist: trusty
+sudo: false
 compiler:
   - clang
   - gcc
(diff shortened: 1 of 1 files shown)


TimWolla:
This is blocked by: https://github.com/travis-ci/apt-package-whitelist/pull/559

Comment if you're generally willing to merge this one, no need to bother the Travis folks unnecessarily.

cperciva:
Looks reasonable to me subject to it working (which it doesn't yet, I assume due to the whitelisting issue you mentioned).

TimWolla:
> which it doesn't yet, I assume due to the whitelisting issue you mentioned

Your assumption is correct: https://travis-ci.org/Tarsnap/spiped/jobs/161430950#L130

They're doing this as there might be possible escape vectors from containers when you're root, as you share the host / kernel. I just asked for an update / review in the whitelist's pull request.

gperciva:
Once #97 is accepted (after #110 and #104), this PR will be unnecessary, as it also switches to the container-based Travis-CI infrastructure.

cperciva:
This can be closed now, right?

gperciva:
Well, we haven't been able to switch yet.  (I switched scrypt, but we can't switch spiped until nmap's `ncat` is replaced)

cperciva:
Oh, I got confused because the other PRs weren't open any more.

gperciva:
Closing this because it's handled by #97.

PR #88, opened 2016-09-27
Remove dubious and dead code in proto_conn.c

TimWolla:
This code is dubious and useless for several reasons:

1. The `(C->t != -1)` condition is useless, as the function already
   would have returned in line 286, if `C->t` was `-1`. Thus this
   part always is `true`.
2. It is not possible for `k_f` and `k_r` to be non-NULL at this place:
   The only place where these struct members are written to is inside
   `proto_conn_create` and `callback_handshake_done`. In the former
   function they always are initialized to NULL, thus the only place
   where a non-NULL value can be written into these struct members is
   inside `callback_handshake_done`. But at that place
   `callback_connect_done` must have returned already (because there is
   no way to perform a handshake, without a connection).

### DIFF
--- a/proto/proto_conn.c
+++ b/proto/proto_conn.c
@@ -291,12 +291,6 @@ callback_connect_done(void * cookie, int t)
 			goto err1;
 	}
 
-	/* If we have connections and keys, start shuttling data. */
-	if ((C->t != -1) && (C->k_f != NULL) && (C->k_r != NULL)) {
-		if (launchpipes(C))
-			goto err1;
-	}
-
 	/* Success! */
 	return (0);
 


TimWolla:
Thinking about it a little more I am not *that* sure about the second bullet point: I am not familiar enough with the code and control flow to decide whether `callback_connect_done` can be called after `callback_handshake_done` on the decrypting side (in case of a slow target). 

In any case: The first bullet point should stand and you should take a look at it at least.

TimWolla:
Sorry for the potentially third email notification …

I'm retracting this pull request in this form, because my reasoning clearly is wrong, after thinking hard about it.

The “useless” check probably was inserted to make the pieces of code in `callback_connect_done` and `callback_handshake_done` identical, right? I propose improving the comments, so that no one else get's confused why this identical piece of code exists twice. My suggestion is something like this:

`callback_connect_done`
```
/* If the handshake already finished: Start shuttling data. */
```
`callback_handshake_done`
```
/* If we already connected to the target: Start shuttling data. */
```

gperciva:
Thanks for examining the code so carefully, @TimWolla!  I've added your suggestion as #89.

gperciva:
I'm closing this PR since the other one is enough to keep track of this suggestion.

PR #89, opened 2016-09-27
Clarify comments about proto_conn launchpipes

gperciva:
This arose from https://github.com/Tarsnap/spiped/pull/88.


cperciva:
I'm fine with the wording change, but changing the commas to colons seems grammatically suspect.

gperciva:
Updated

PR #90, opened 2016-10-05
Valgrind suppressions

gperciva:


gperciva:
This uses the same framework as https://github.com/Tarsnap/scrypt/pull/66, but doesn't apply it to the actual `test_spiped.sh` script yet.  (that's coming in a separate PR)

Note the addition of `NOINST = 1` to the `Makefile.BSD` file in a program subdirectory.  Technically this isn't required, because I didn't add `${TEST}` to the `install:` section of the top-level `Makefile`, but I thought it was worth adding.

gperciva:
This build failure is due to travis-CI changing their OSX build image: https://blog.travis-ci.com/2016-10-04-osx-73-default-image-live/

The actual error given is:
```
../libcperciva/crypto/crypto_aes.c:6:25: fatal error: openssl/aes.h: No such file or directory
 #include <openssl/aes.h>
                         ^
compilation terminated.
```

As you can see, this means that we've run straight into OSX's removal (instead of mere deprecation) of libssl in OSX 11.  I'm preparing a fix.

cperciva:
> Note the addition of NOINST = 1 to the Makefile.BSD file in a program subdirectory. Technically this isn't required, because I didn't add ${TEST} to the install: section of the top-level Makefile, but I thought it was worth adding.

I'm not a fan of adding a new Makefile.prog-noinst; instead, how about adding some logic to the install target in Makefile.prog which checks NOINST and errors out?  Something like

```
install:${PROG}
        if ! [ -z "${NOINST}" ]; then
                echo "${NOINST}"
                exit 1
        fi
        mkdir -p ${BINDIR}
etc etc etc.
```

And then of course add NOINST to the set of variables we copy from Makefile.BSD into Makefile, and set it to "This is test code not intended to be installed" or something similar.

gperciva:
But then we have junk in `Makefile` that we don't need!  To simplify things, how about we use perl?  ;)

(once you finish cringing, see the revised first commit in this PR)

Makefile.BSD
```diff
@@ -14,7 +14,14 @@ ${D}/Makefile::
 	( cd ${D} && printf 'SRCS=' && ${MAKE} -V SRCS | sed -e 's| cpusupport-config.h||' ) >> $@
 	( cd ${D} && printf 'IDIRS=' && ${MAKE} -V IDIRS ) >> $@
 	( cd ${D} && printf 'LDADD_REQ=' && ${MAKE} -V LDADD_REQ ) >> $@
-	cat Makefile.prog >> $@
+	# Copy the Makefile template, removing the "install:"
```
cperciva: Normally I like comments... but makefile parsers are finicky things, and I would be completely unsurprised if a comment here caused problems for some versions of make.

Actually, I *think* this would simply get passed through to the shell, which would interpret it as a comment... but still, let's not take chances.

gperciva: You know this is in `Makefile.BSD`, right?  So it's only getting run by BSD make on either your computer or mine?  :)

Still, updated PR accordingly.  Once this is accepted, I'll send in another PR that moves the other comments in this file outside of the rules.  (I don't want to do this now, because that would cause merge conflicts between the two PRs, so I'd have to edit it later anyway)

PR #94, opened 2016-10-15
build: fix "if ... NOINST" line

gperciva:
The earlier line worked by coincidence: when there was no NOINST, it would display the error:

    /bin/sh: 1: [: -eq: unexpected operator

and then take the "else" command.  The end result was the Makefiles that we desired, but we didn't receive the result in an appealing manner.


gperciva:
Sorry, I missed this error due to too many other things happening on the command-line.  I've changed my "rebuild makefiles" script to `>/dev/null` so that I can see any important stuff.

cperciva:
TIL about the zero-argument form of the `[` command.

PR #96, opened 2016-11-08
Tests multiple

gperciva:


gperciva:
This assumes that PR https://github.com/Tarsnap/scrypt/pull/84 (comparing results with the system's scrypt) is accepted, so shouldn't be merged until then.

gperciva:
Updated PR

PR #97, opened 2016-11-17
Tests nmap replace

gperciva:


gperciva:
First draft; would allow #87 container-based Travis-CI.  (although if we want valgrind testing, that would negate this possibility as they haven't whitelisted valgrind: https://github.com/travis-ci/apt-package-whitelist/issues/3635)

Still, it would be nice not to have the extra dependency for the test suite.

tests/spiped_nc/graceful_shutdown.h
```diff
@@ -0,0 +1,18 @@
+#ifndef GRACEFUL_SHUTDOWN_H
+#define GRACEFUL_SHUTDOWN_H
+
+/**
+ * graceful_shutdown_register(callback, caller_cookie):
+ * Initializes an event which checks for SIGTERM every 1.0
```
cperciva: Why is this line wrapping so early?

tests/spiped_nc/graceful_shutdown.c
```diff
@@ -0,0 +1,96 @@
+#include <signal.h>
+#include <stdlib.h>
+
+#include "events.h"
+#include "warnp.h"
+
+#include "graceful_shutdown.h"
+
+struct graceful_shutdown_cookie {
+	void (* begin_shutdown)(void *);
```
cperciva: I would have called these `callback` and `cookie` personally.  Not a big deal, but what does the verbosity get you?

cperciva: Also, we usually have callbacks return `int`, so that they can report if something goes fatally wrong.

tests/spiped_nc/graceful_shutdown.c
```diff
@@ -0,0 +1,96 @@
...
+graceful_shutdown(void * cookie)
+{
+	struct graceful_shutdown_cookie * G = cookie;
+
+	/* This timer has expired. */
+	graceful_shutdown_timer_cookie = NULL;
+
+	/* Use the callback function, or schedule another check in 1 second. */
+	if (should_shutdown)
+		G->begin_shutdown(G->caller_cookie);
```
cperciva: Where is `G` freed?

tests/spiped_nc/graceful_shutdown.c
```diff
@@ -0,0 +1,96 @@
...
+
+#include "graceful_shutdown.h"
+
+struct graceful_shutdown_cookie {
+	void (* begin_shutdown)(void *);
+	void * caller_cookie;
+};
+
+static volatile sig_atomic_t should_shutdown = 0;
+static void * graceful_shutdown_timer_cookie = NULL;
```
cperciva: Why have this here, but the cookie and callback as part of a structure which gets allocated on the heap, anyway?

tests/spiped_nc/graceful_shutdown.c
```diff
@@ -0,0 +1,96 @@
...
+	struct graceful_shutdown_cookie * G = cookie;
+
+	/* This timer has expired. */
+	graceful_shutdown_timer_cookie = NULL;
+
+	/* Use the callback function, or schedule another check in 1 second. */
+	if (should_shutdown)
+		G->begin_shutdown(G->caller_cookie);
+	else
+		graceful_shutdown_timer_cookie = events_timer_register_double(
```
cperciva: If this fails, we should return -1.

tests/spiped_nc/graceful_shutdown.c
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * graceful_shutdown_shutdown(cookie):
+ * Stop and free the graceful shutdown timer.
+ */
+void
+graceful_shutdown_shutdown(void * cookie)
+{
+
+	if (graceful_shutdown_timer_cookie != NULL)
```
cperciva: Restoring the old signal handler would be a polite thing to do.  Of course, you would need to remember what it was.

tests/spiped_nc/main.c
```diff
@@ -0,0 +1,128 @@
...
+#include "graceful_shutdown.h"
+#include "simple_server.h"
+
+#define MAX_CONNECTIONS 2
+
+struct nc_cookie {
+	FILE * out;
+};
+
+/* Forward definition. */
```
cperciva: Why not static?

tests/spiped_nc/main.c
```diff
@@ -0,0 +1,128 @@
...
+	uint16_t port;
+	const char * filename;
+
+	(void) argc; /* UNUSED */
+
+	WARNP_INIT;
+
+	/* Parse command-line arguments. */
+	if (argc < 3) {
+		fprintf(stderr, "%s PORT FILENAME\n", argv[0]);
```
cperciva: `usage:`

tests/spiped_nc/simple_server.c
```diff
@@ -0,0 +1,363 @@
...
+	int (* callback_nc_message)(void *, uint8_t *, size_t);
+};
+
+/* Doubly linked list. */
+struct conn_list_node {
+	/* General "dispatch"-level info. */
+	struct conn_list_node * prev;
+	struct conn_list_node * next;
+	struct accept_state * A;
+	/* Reading a network message. */
```
cperciva: I would separate the sections with a blank line her.e

tests/spiped_nc/graceful_shutdown.c
gperciva: As for `cookie`... you'd like to see this?

struct cookie {
        int (* begin_shutdown)(void *);
        void * cookie;
};

Having two different `cookie` names seemed a bit... "off".

gperciva: Also, looking at `graceful_shutdown.h`:

```
void * graceful_shutdown_register(callback, caller_cookie);
void graceful_shutdown_shutdown(cookie);
```

Wouldn't it be nice to distinguish between the "graceful_shutdown cookie", and "the cookie we pass to the shutdown callback" ?

tests/spiped_nc/graceful_shutdown.c
gperciva: `G` is freed in `graceful_shutdown_shutdown()`, which is called by `main.c` in the success and failure (`err4`) cases.

gperciva:
Updated PR; changed everything you mentioned other than keeping `name_cookie` so we can avoid:

```
struct cookie {
        ...
        void * cookie;
};
```

If it looks good, I'll rewrite history.

tests/spiped_nc/graceful_shutdown.c
```diff
@@ -0,0 +1,122 @@
...
+	return (-1);
+}
+
+/**
+ * graceful_shutdown_register(callback, caller_cookie):
+ * Initializes an event which checks for SIGTERM every 1.0 seconds; if
+ * detected, it calls the ${callback} function and passes it the ${cookie}.
+ */
+void *
+graceful_shutdown_register(int (* begin_shutdown)(void *),
```
cperciva: Is it your intention that more than one graceful shutdown callback might be registered at once?

If yes, you need to make sure you restore signal handlers in the right order.

If no, why bother returning a cookie when you could just keep all the parameters as static variables inside this file?

gperciva:
Closing this until we get #104 added.  I'm now investigating the OSX Travis-CI flakiness in more detail.

gperciva:
This includes #110 and #104; see those before looking at this.  I'm just re-opening this PR so that we can keep track of this, and to schedule the Travis-CI build.

gperciva:
I'm sure that this won't get in tonight, but at least it's open now so it's easier to keep "in mind".

gperciva:
Updated this PR to remove code that you'd probably ask me to remove anyway.  (i.e. `simple_server` now no longer has `separate _init()`, `_run()`, `_shutdown()`; now it's just a single function).

tests/nc-server/simple_server.h
```diff
@@ -0,0 +1,16 @@
...
+#include <stdint.h>
+
+/**
+ * simple_server_init(port, nconn_max, shutdown_after, callback, caller_cookie):
+ * Run a server which accepts up to ${nconn_max} connections to port ${port}.
+ * After receiving a message, call ${callback} and pass it the
+ * ${caller_cookie}, along with the message.  Automatically shuts itself down
+ * after ${shutdown_after} connections have been dropped.
+ */
+int simple_server(const uint16_t, const size_t, const size_t,
```
cperciva: `port` should be a `in_port_t`.  But why not simply pass the entire address string in?

tests/nc-server/simple_server.h
```diff
@@ -0,0 +1,16 @@
+#ifndef SIMPLE_SERVER_H
+#define SIMPLE_SERVER_H
+
+#include <stdint.h>
+
+/**
+ * simple_server_init(port, nconn_max, shutdown_after, callback, caller_cookie):
```
cperciva: _init ?

tests/nc-server/simple_server.c
```diff
@@ -0,0 +1,368 @@
...
+err0:
+	/* Failure! */
+	return (-1);
+}
+
+/**
+ * simple_server_shutdown(cookie):
+ * Stop and free memory associated with the ${cookie}.
+ */
+void
```
cperciva: This can now be static, right?

tests/nc-server/main.c
```diff
@@ -0,0 +1,106 @@
...
+
+err0:
+	/* Failure! */
+	return (-1);
+}
+
+int
+main(int argc, char ** argv)
+{
+	struct nc_cookie * C;
```
cperciva: Why not simply make this a structure on the stack; what do you gain by making this a pointer and the allocating the structure?

tests/nc-server/main.c
```diff
@@ -0,0 +1,106 @@
...
+	unsigned long port_long;
+	uint16_t port;
+	const char * filename;
+
+	(void)argc; /* UNUSED */
```
cperciva: Not unused.

tests/nc-server/main.c
```diff
@@ -0,0 +1,106 @@
...
+	filename = argv[2];
+
+	/* Initialize our cookie. */
+	if ((C = malloc(sizeof(struct nc_cookie))) == NULL) {
+		warn0("Out of memory.");
+		goto err0;
+	}
+
+	/* Open the output file; can be /dev/null. */
+	if ((C->out = fopen(filename, "w+b")) == NULL) {
```
cperciva: Why `w+b` rather than simply `wb`?  Are you expecting to read from this file?

tests/nc-server/simple_server.h
cperciva: BTW the consts are meaningless here.  These are pass-by-value so the function can't modify the original variables anyway.

tests/nc-server/simple_server.c
```diff
@@ -0,0 +1,368 @@
...
+		goto err0;
+	}
+	node_new->prev = NULL;
+	node_new->next = NULL;
+	node_new->A = A;
+	node_new->sock_read = s;
+
+	/* Schedule reading from this connection. */
+	if ((node_new->network_read_cookie = network_read(node_new->sock_read,
+	    node_new->buf, BUFLEN, 0, callback_read, node_new)) == NULL) {
```
cperciva: You should ask for at least 1 byte.

tests/nc-server/simple_server.c
```diff
@@ -0,0 +1,368 @@
...
+	A->accept_cookie = NULL;
+
+	/* If we got a -1 descriptor, something went seriously wrong. */
+	if (s == -1) {
+		warnp("network_accept");
+		goto err0;
+	}
+
+	/* We have gained a connection. */
+	A->nconn += 1;
```
cperciva: Maybe this should be rolled back in the error path if we fail to allocate a node or start reading from the socket?

tests/nc-server/simple_server.c
```diff
@@ -0,0 +1,368 @@
...
+	/* Cookie is no longer valid. */
+	R->network_read_cookie = NULL;
+
+	/* If we have a message. */
+	if (lenread > 0) {
+		/* Handle it with the parent code. */
+		A->callback_nc_message(A->caller_cookie, R->buf,
+		    (size_t)lenread);
+
+		/* We can re-use R. */
```
cperciva: `/* Try to read some more data. */` might be clearer?

tests/nc-server/simple_server.h
gperciva: Passing in the entire address string: I'd rather do the command-line parsing in the command-line parsing area.  If I passed in argv[1], then if they mistyped a port number, the error would come from `simple_server()`, not the beginning of `int main()`.

gperciva:
Updated PR

tests/nc-server/simple_server.c
```diff
@@ -0,0 +1,372 @@
...
+		goto err0;
+	}
+
+	/* Success! */
+	return (0);
+
+err2:
+	free(node_new);
+err1:
+	A->nconn += 1;
```
cperciva: Umm, what?

tests/nc-server/simple_server.c
```diff
@@ -0,0 +1,372 @@
...
+	sock_addr_freelist(sas);
+err1:
+	free(addr);
+	events_shutdown();
+err0:
+	/* Failure! */
+	return (-1);
+}
+
+/**
```
cperciva: Static functions go before non-static functions unless there's a very good reason to do otherwise.

cperciva:
Looks reasonable.  Please tidy up the commit history.

gperciva:
Updated PR with rebased history

cperciva:
Looks good.  I'll merge when I notice that the tests have passed...

PR #100, opened 2016-11-22
Avoid obscure English grammar

gperciva:
Yes, people who did a degree in Britain will know that "writ" is the archaic past participle of "write".  But y'know what?  Even you didn't use "writ" consistently!:

    ${callback}(${cookie}, lenwrit), where lenwrit is -1 on
    error and the number of bytes written... otherwise.
                                      ~~~

Let's standardize on "written", and thus use "lenwritten".


gperciva:
While we're looking at this spot, should the second `lenwritten` have any punctuation around it?  I'm not certain how you normally refer to a callback argument, as opposed to a function `${argument}`.

gperciva:
I think this is a timing issue with the tests, but I'd like to get #97 (which replaces `nmap` with our `spiped_nc`) in before I try fiddling with stuff.

cperciva:
Nope.  I'm an archaic sort of guy, and I value conciseness in my variable names.

PR #102, opened 2016-12-19
Typo: unix -> Unix

gperciva:


libcperciva/util/sock.c
```diff
@@ -233,7 +233,7 @@ sock_resolve(const char * addr)
 	char * ips;
 	long p;
 
-	/* If the address starts with '/', it's a unix socket. */
+	/* If the address starts with '/', it's a Unix socket. */
```
cperciva: Hmm, to be perfectly accurate this should probably be "Unix domain socket".

gperciva:
Updated PR

PR #104, opened 2017-01-08
Graceful shutdown

gperciva:


gperciva:
This is split off from https://github.com/Tarsnap/spiped/pull/97

gperciva:
hmm, works in FreeBSD.  I'll reboot into OSX this afternoon and check if it's just a travis-CI load issue or a real thing.

lib/util/graceful_shutdown.c
```diff
@@ -0,0 +1,128 @@
...
+{
+
+	/* Start signal handler and save the original one. */
+	if ((sighandler_sigterm_orig = signal(SIGTERM,
+	    graceful_shutdown_handler)) == SIG_ERR) {
+		warnp("signal");
+		goto err0;
+	}
+
+	/* Success ! */
```
TimWolla: Additional space before the exclamation mark.

lib/util/graceful_shutdown.c
```diff
@@ -0,0 +1,128 @@
...
+	caller_cookie = caller_cookie_parent;
+
+	/* Periodically check whether a signal was received. */
+	if ((timer_cookie = events_timer_register_double(
+	    graceful_shutdown, NULL, 1.0)) == NULL) {
+		warnp("Failed to register graceful shutdown timer");
+		goto err0;
+	}
+
+	/* Success ! */
```
TimWolla: Additional space before the exclamation mark.

gperciva: thanks, fixed these two

gperciva:
Trigger new test build; it worked on osx manually, so I suspect the previous problem was due to Travis-CI load on their osx machines.  (I still haven't managed to reproduce this manually on osx)

gperciva:
This PR does not contain the "print extra debug info" and, more importantly, the "don't use hard-coded `sleep(1)` values" that are in PR #107 .  So I'm not terribly disappointed (or surprised) by this build failure.  In fact, that gives me more faith in #107 -- I've tried it more than 6 times in the past few days without problems, whereas this one died instantly.

Once the other PR has been merged, I'll trigger a new build here.

gperciva:
This now includes PR #110.

lib/util/graceful_shutdown.c
```diff
@@ -0,0 +1,128 @@
...
+	return (0);
+
+err0:
+	/* Failure! */
+	return (-1);
+}
+
+/* Stop and free the graceful shutdown timer. */
```
cperciva: Given that this only gets called from one place, why not include it in that function?

lib/util/graceful_shutdown.c
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * graceful_shutdown_initialize():
+ * Initializes a signal handler for SIGTERM.
+ */
+int
+graceful_shutdown_initialize()
+{
+
+	/* Start signal handler and save the original one. */
```
cperciva: Why is this a separate function?  Why not just initialize when graceful_shutdown_monitor gets called?

gperciva: I was uncertain about how this would play out in spiped, particularly with regards to the docker container workaround.  One of my overall goals in crafting `graceful_shutdown.h` was to minimize the diff in `spiped/main.c`, and that resulted in splitting the functions in that way.

cperciva: The docker workaround is for SIGINT.  This is SIGTERM.  There's no interaction at all between them.

spiped/main.c
```diff
@@ -336,9 +308,12 @@ main(int argc, char * argv[])
...
 	}
 
-	/* Check periodically whether a signal was received. */
-	graceful_shutdown_timer_cookie = events_timer_register_double(
-	    graceful_shutdown, dispatch_cookie, 1.0);
+	/* Register a handler for SIGTERM. */
+	if (graceful_shutdown_initialize(&callback_graceful_shutdown,
+	    dispatch_cookie)) {
+		warn0("Failed to start graceful_shutdown timer");
+		goto err5;
```
cperciva: err6, I think?

lib/util/graceful_shutdown.c
cperciva: Was this fixed?  Because there seems to be an extra space here again...

lib/util/graceful_shutdown.c
```diff
@@ -0,0 +1,105 @@
...
+
+	/* Use the callback function, or schedule another check in 1 second. */
+	if (should_shutdown) {
+		if (begin_shutdown(caller_cookie) != 0) {
+			warn0("Failed to begin shutdown");
+			goto err0;
+		}
+
+		/* Cancel timer. */
+		if (timer_cookie != NULL)
```
cperciva: Umm, really?

lib/util/graceful_shutdown.c
gperciva: I thought I had... oh well.  `git grep "Success "` now returns nothing.

gperciva:
re: all these failing tests: at least the good news is that this [should be] the last PR that has this randomness.  The end result of the nmap replacement PR should produce reliable tests, or at least an exact error message about what failed.

lib/util/graceful_shutdown.c
```diff
@@ -0,0 +1,101 @@
...
+
+#include "graceful_shutdown.h"
+
+/* Data from parent code. */
+static int (* begin_shutdown)(void *);
+static void (* sighandler_sigterm_orig)(int);
+static void * caller_cookie;
+static void * timer_cookie;
+
+/* Flag to show that SIGTERM was given. */
```
cperciva: s/given/received/

lib/util/graceful_shutdown.c
```diff
@@ -0,0 +1,101 @@
...
+static volatile sig_atomic_t should_shutdown = 0;
+
+/* Forward definitions. */
```
cperciva: Don't think either of these are needed?

cperciva:
Two very minor nits.  Please rebase, then I'll merge.

lib/util/graceful_shutdown.c
gperciva: I'd argue that we don't care about *receiving* SIGTERM; what we care about was that it was sent to us.  Which implicitly means that we received it, but ultimately that's not what we care about.  However, I'd much rather get this merged, so I'll refrain from making that argument.  :)

cperciva: What we *care about* is whether it was sent.  But what we *know about* is whether it was received. :-)

gperciva:
oops, it needs stddef.h.  Wait a moment

gperciva:
Fascinating!  The previous build test failed with:
```
  05-send-data-spiped... nc-client: Mismatch between data sent and data requested to send
```
https://travis-ci.org/Tarsnap/spiped/jobs/203329571

and that's something we added as a theoretical concern: https://github.com/Tarsnap/spiped/pull/110#discussion_r101208380
[ok, that comment doesn't sound terribly "theoretical", but my recollection of us walking to skytrain for the dress rehearsal was that this case wasn't expected to occur, but it would be good practice (in both meanings of the term) to add it anyway.]

I'll go ahead with the graceful-shutdown PR anyway, but I need to rethink how the tests function, because I has no idea it could fail in that manner.

gperciva:
... and yes, this PR has been updated, and just passed the tests.  Good for merging now.

cperciva:
Ah, that error message isn't actually correct.  The network_write callback should be invoked with one of the values { -1, [minlen, maxlen] }; since minlen = maxlen in this case, getting the wrong number of bytes means that we have lenwrit == -1, aka. "an error occurred".

So the /* Check that we wrote all our data. */ block should (in the case that lenwrit != send->nchars) assert(lenwrit == -1) and then warnp("Failed to send data");

cperciva:
(But you can do that in a future PR; I'll merge this one now.)

PR #105, opened 2017-01-13
Silence scan-build warning about argv

gperciva:


cperciva:
What's the warning?  This seems like an odd place for scan-build to be confused.

gperciva:
```
main.c:225:2: warning: Value stored to 'argv' is never read
        argv += optind;
        ^       ~~~~~~
```

I've rebased on this PR on top of #107 (hoping to get another Travis-CI build failure so I can get more debug info).

cperciva:
> main.c:225:2: warning: Value stored to 'argv' is never read

Silly analyzer, doesn't it know that argc/argv are idiomatic?

I'm fine with adding a `(void)argv;` to silence this, but let's have a more meaningful comment; say, `/* argv is not used beyond this point. */`.

PR #106, opened 2017-01-13
Test debug failure

gperciva:


gperciva:
I added these commits to attempt to get more info about why the spiped builds were failing, but so far it hasn't failed.  This could be random chance, or it could be due to giving the `[pid]` to `wait`.  Either way, I think it's worth merging it.

PR #104 (graceful shutdown) now includes these commits.

gperciva:
yay, got a failure again.  Time for the big guns: `sh -x`.

PR #107, opened 2017-01-15
Improve tests' reliability and verbosity

gperciva:


gperciva:
This should be more robust; at the very least, it gives more info in most failure cases.  It seems like the travis-CI servers aren't too busy right now, so these changes aren't tested under heavy load.

gperciva:
oooakay.  I'm not surprised that the test failed because I claimed it should be more robust; but I *am* rather surprised at how it failed.  Much to think about now!

gperciva:
Ok, my latest theory is that under heavy loads (particularly heavy *disk* loads), `nc` gets `kill`ed before it finishes writing the output file.  How plausible this is depends on the Travis-CI setup: their hardware, virtualization software, etc.

However, it doesn't much matter at this point.  The #97 nmap replacement takes care of this -- even without seeing these errors, I was paranoid about catching `SIGTERM` and writing the buffer to a file before quitting.  So I think the best thing to do is to merge this (unless there's stylistic problems), then re-evaluate #104 and #97.
(and if/when the problems due to load re-emerge, the `VERBOSE` info that this PR generates will give us more evidence for or against my theory)

gperciva:
Ok, after more than half a dozen successful builds, I (partially) give up on seeing detailed info about a failed build.  I've removed the `FIXME` commits from this PR, so it now contains commits which I believe to be good.  In particular, it will print a bit more info if there *is* a test failure, and it also removes hard-coded `sleep(1)` values.

If and when we start seeing build failures again, I can enable `set -x` and try to get more info.

PR #108, opened 2017-01-22
Poll

gperciva:


cperciva:
Doesn't this want Makefiles regenerated to add `ctassert.h` as a dependency for `events_network.o`?

Also, I have no idea why Travis-CI failed...

gperciva:
This includes #109 "tests simplification", but I think it's fine to go ahead and merge it.  Right now, Travis-CI has a backlog of over 1000 OSX builds, and a partial DNS outage.  No point waiting for that to complete before examining this and merging it.

(As for graceful-shutdown, the latest build failure is because the test organizer process lost contact with one of the OSX builds; this occasionally happens with high loads and/or DNS failures, so I'm not counting that as a mark against the simplified tests.)

PR #110, opened 2017-02-06
nc-client

gperciva:


tests/test_spiped.sh
```diff
@@ -106,7 +106,7 @@ setup_spiped_encryption_server () {
 		-s [127.0.0.1]:${src_port}	\
 		-t [127.0.0.1]:${mid_port}	\
 		-p ${s_basename}-spiped-e.pid	\
-		-k /dev/null -o 1
+		-D -k /dev/null -o 1
```
cperciva: This is weird, and does exactly the opposite of what I think you're trying to do.

tests/nc-client/main.c
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	/* Connect to target. */
+	if ((connect_cookie = network_connect(sas_t, callback_connected,
+	    send)) == NULL) {
+		warn0("Error connecting");
+		goto err2;
+	}
+
+	/* Loop until we die. */
```
cperciva: This is explicitly not looping until we die.  Rather, it's looping until we're finished with the connection.

tests/nc-client/main.c
```diff
@@ -0,0 +1,151 @@
...
+	if (send->buffer != NULL) {
+		free(send->buffer);
+		send->buffer = NULL;
+	}
+
+	/* Success! */
+	exit(0);
+
+err3:
+	network_connect_cancel(connect_cookie);
```
cperciva: We might arrive here after the connect callback has been invoked.

tests/test_spiped.sh
gperciva: huh.

tests/nc-client/main.c
```diff
@@ -0,0 +1,151 @@
...
+
+	/* Read data from stdin. */
+	if ((read = getline(&send->buffer, &len, stdin)) != -1) {
+		/* Send data to server. */
+		if (network_write(send->socket, (uint8_t *)send->buffer,
+		    (size_t)read, (size_t)read, callback_wrote, cookie)
+		    == NULL) {
+			warn0("network_write failure");
+			goto err0;
+		};
```
cperciva: `;`

tests/nc-client/main.c
```diff
@@ -0,0 +1,151 @@
...
+
+/* Send data from stdin to a socket, or close the connection. */
+static int
+send_input(void * cookie)
+{
+	struct senddata * send = (struct senddata *)cookie;
+	size_t len = 0;
+	ssize_t read;
+
+	/* Ensure that the buffer is clear. */
```
cperciva: Since the buffer is guaranteed to be NULL when we first arrive here, why not put the buffer-freeing into callback_wrote?

gperciva:
Technically, I didn't need to keep the `write_cookie` -- the only way [1] we'd exit out of the event loop is from `send_input()`, which never has an existing write cookie.  But I figured it was good practice for my knowledge of the libcperciva network API to add it.  :)

[1] other than SIGTERM / SIGKILL, which I'm not going to bother with.

cperciva:
On 02/14/17 20:50, Graham Percival wrote:
> Technically, I didn't need to keep the |write_cookie| -- the only way [1] we'd
> exit out of the event loop is from |send_input()|, which never has an existing
> write cookie.

You sure about that?  What if poll(2) fails for some reason?

-- 
Colin Percival
Security Officer Emeritus, FreeBSD | The power to serve
Founder, Tarsnap | www.tarsnap.com | Online backups for the truly paranoid

gperciva:
huh, yes, I suppose that the underlying API could have a problem.  Good lesson!

gperciva:
if this looks good, I'll rebase the history.

tests/nc-client/main.c
```diff
@@ -0,0 +1,157 @@
...
+	return (1);
+}
+
+/* Finished writing data; look for more from stdin. */
+static int
+callback_wrote(void * cookie, ssize_t lenwrit)
+{
+	struct senddata * send = (struct senddata *)cookie;
+
+	(void)lenwrit; /* UNUSED */
```
cperciva: Maybe you want to check `lenwrit`?  If the destination process closes the socket before all the data is sent, do you want nc_client to report an error?

cperciva:
Looks fine aside from the one question about `lenwrit`.

gperciva:
Updated PR with new error-checking.  And yes, I tried sending my freebsd-i386 `.vdi` to ncat, to ensure that I could kill it during a transmission and get an appropriate error message.  (and have no valgrind errors)

I rebased the previous REBASE commit, but added two more.

spipe/main.c
```diff
@@ -176,7 +176,7 @@ main(int argc, char * argv[])
 		goto err3;
 	}
 
-	/* Loop until we die. */
+	/* Loop until we're done with the connection. */
```
cperciva: Fine, but it doesn't belong in this PR.

tests/nc-client/main.c
```diff
@@ -0,0 +1,175 @@
...
+	char * buffer;
+	ssize_t nchars;
+	int socket;
+	int conndone;
+	void * connect_cookie;
+	void * write_cookie;
+};
+
+/* Forward declarations. */
+static int send_input(void *);
```
cperciva: Not necessary, since the function is defined before it's first used.  I think it's only `callback_wrote` which needs a forward declaration.

tests/nc-client/main.c
```diff
@@ -0,0 +1,175 @@
...
+static int callback_wrote(void * cookie, ssize_t lenwrit);
```
cperciva: And you don't need variable names here.

tests/nc-client/main.c
```diff
@@ -0,0 +1,175 @@
...
+	if ((send->nchars = getline(&send->buffer, &len, stdin)) != -1) {
+		/* Send data to server. */
+		if ((send->write_cookie = network_write(send->socket,
+		    (uint8_t *)send->buffer, (size_t)send->nchars,
+		    (size_t)send->nchars, callback_wrote, cookie)) == NULL) {
+			warn0("network_write failure");
+			goto err0;
+		}
+	} else {
+		/* Close connection without sending anything. */
```
cperciva: You might want to check `feof(stdin)` here, to distinguish between EOF and an error?

tests/nc-client/main.c
```diff
@@ -0,0 +1,175 @@
...
+callback_wrote(void * cookie, ssize_t lenwrit)
+{
+	struct senddata * send = (struct senddata *)cookie;
+
+	/* We are not longer writing. */
+	send->write_cookie = NULL;
+
+	/* Check that we wrote all our data. */
+	if (lenwrit != send->nchars) {
+		warn0("Mismatch in data sent and data requested to send");
```
cperciva: s/in/between/ ?

tests/nc-client/main.c
```diff
@@ -0,0 +1,175 @@
...
+	return (1);
+}
+
+/* Got a connection; look for data from stdin. */
+static int
+callback_connected(void * cookie, int socket)
+{
+	struct senddata * send = (struct senddata *)cookie;
+
+	/* We are not longer connecting. */
```
cperciva: s/not longer/no longer/

tests/nc-client/main.c
```diff
@@ -0,0 +1,175 @@
...
+		/* Close connection without sending anything. */
+		send->conndone = 1;
+	}
+
+	/* Success! */
+	return (0);
+
+err0:
+	/* Failure! */
+	return (1);
```
cperciva: We normally return -1 on failure, not 1.  (Although it doesn't really matter.)

tests/nc-client/main.c
```diff
@@ -0,0 +1,175 @@
...
+
+	/* Clear the buffer that was used. */
+	free(send->buffer);
+	send->buffer = NULL;
+
+	return send_input(cookie);
+
+err0:
+	/* Failure! */
+	return (1);
```
cperciva: We normally return -1 on failure, not 1.  (Although it doesn't really matter.)

tests/nc-client/main.c
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	/* Record socket for future use. */
+	send->socket = socket;
+
+	return send_input(cookie);
+
+err0:
+	/* Failure! */
+	return (1);
```
cperciva: We normally return -1 on failure, not 1.  (Although it doesn't really matter.)

cperciva:
Looks good aside from the nit-picking.

tests/nc-client/main.c
gperciva: Whoops, I was thinking about error codes.  Should the `return (1)` in `proto_crypt_dh_validate()` be changed?

cperciva: No, `proto_crypt_dh_validate` isn't returning `1` to mean "an internal error occurred"; it's successfully saying "this is not valid".  Think "true / false / file not found" logic states.

gperciva:
Updated PR: rebased the previous updates, added more nitpick-fixing, and removed the unaffiliated commit.

cperciva:
Looks good.  Rebase and I'll merge once the checks finish.

gperciva:
Rebased and tests ok.

gperciva:
wait, I just realized it needs another `warn0`.

gperciva:
ok, actually ready now.

PR #114, opened 2017-02-20
tests/nc-client: clarify client network errors, block on socket close

gperciva:
Yes, given the specific values inside this file's send_input() function,
we're not expecting to see a lenwrit which is neither -1 nor send->nchars.
However, I think the code is more clear this way, and it's not unlikely that
I'll end up copying bits of this file the next time I want to send data over a
network.  Also, if we did somehow see an unexpected value for lenwrit, I think
it would be nice to have a specific error message for that case.

### DIFF
--- a/tests/nc-client/main.c
+++ b/tests/nc-client/main.c
@@ -1,3 +1,4 @@
+#include <assert.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -64,10 +65,13 @@ callback_wrote(void * cookie, ssize_t lenwrit)
 	/* We are no longer writing. */
 	send->write_cookie = NULL;
 
-	/* Check that we wrote all our data. */
-	if (lenwrit != send->nchars) {
-		warn0("Mismatch between data sent and data requested to send");
+	/* Check results. */
+	if (lenwrit == -1) {
+		warnp("network_write send");
 		goto err0;
+	} else {
+		/* We should have sent everything. */
+		assert(lenwrit == send->nchars);
 	}
 
 	/* Clear the buffer that was used. */
--- a/tests/01-connection-open-close-single.sh
+++ b/tests/01-connection-open-close-single.sh
@@ -13,11 +13,12 @@
 scenario_cmd() {
 	# Set up infrastructure.
 	setup_spiped_decryption_server
+	setup_spiped_encryption_server
 
 	# Open and close a connection.
 	setup_check_variables
 	(
-		echo "" | ${nc_client_binary} [127.0.0.1]:${mid_port}
+		echo "" | ${nc_client_binary} [127.0.0.1]:${src_port}
 		echo $? > ${c_exitfile}
 	)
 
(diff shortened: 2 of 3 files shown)


gperciva:
ok, I'm going to bed soon.

tests/nc-client/main.c
```diff
@@ -64,9 +64,13 @@ callback_wrote(void * cookie, ssize_t lenwrit)
...
 
-	/* Check that we wrote all our data. */
-	if (lenwrit != send->nchars) {
-		warn0("Mismatch between data sent and data requested to send");
+	/* Check for failure cases. */
+	if (lenwrit == -1) {
+		warnp("network_write send");
+		goto err0;
+	} else if (lenwrit != send->nchars) {
+		warnp("Mismatch between data sent: %zu, and data "
```
cperciva: It doesn't matter all that much in this case, but this really should be an assert, since the valid values passed to the network_write callback are specified by the API.  Warnings are for "something external failed" (we ran out of memory, there was a network error, etc.), but we use assertions for "we have observed a bug in our code" since that triggers a core dump (which is useful for figuring out why our code is broken).

gperciva:
ok, updated, rebased, ready for merge if you feel like it.

gperciva:
Looks like the complete solution for the "blocking nc-client":
- client shutdown
- server close
- client read
- client quit
does work; I tried three builds with everything together, and none of them failed.

The bad news is that if we want to have logical, self-contained PRs, we need to slog through a few more rounds of flaky tests.  The new client is not compatible with the old ncat server, while the new server (alone) still has that 10-15% random failure.

The current two PRs should function independently:
- this one, doing minor cleanup of of nc-client
- #97 , which adds nc-server (and which will probably require a few rounds of redesign)

tests/nc-client/main.c
```diff
@@ -64,11 +65,12 @@ callback_wrote(void * cookie, ssize_t lenwrit)
 	/* We are no longer writing. */
 	send->write_cookie = NULL;
 
-	/* Check that we wrote all our data. */
-	if (lenwrit != send->nchars) {
-		warn0("Mismatch between data sent and data requested to send");
+	/* Check for failure cases. */
```
cperciva: Maybe change the second paragraph of the commit message for this commit?

tests/nc-client/main.c
```diff
@@ -101,6 +103,17 @@ callback_connected(void * cookie, int socket)
...
+	if ((opts = fcntl(send->socket, F_GETFL, 0)) == -1) {
+		warnp("fcntl get");
+		goto err0;
+	}
+	opts = opts & (~O_NONBLOCK);
+	if (fcntl(send->socket, F_SETFL, opts) == -1) {
+		warnp("fcntl set");
+		goto err0;
+	}
+
```
cperciva: I'm confused.  Why do you want blocking networking?

... if you *do* want blocking networking, you would need to stop using all the rest of the libcperciva networking code, since it's designed to deal with non-blocking sockets...

gperciva: Hmm, I added that while I was trying to figure out the latest set of test failures, and it didn't hurt anything.  I've removed it now.

gperciva:
Revised comments, commit messages, and removed the blocking socket.

PR #116, opened 2017-02-28
tests/nc-client: shutdown when the socket is closed

gperciva:
This makes nc-client block on its shutdown message being received by the server.


tests/nc-client/main.c
```diff
@@ -16,11 +18,40 @@ struct senddata {
 	int conndone;
 	void * connect_cookie;
 	void * write_cookie;
+	void * read_cookie;
+	uint8_t shutdown;
```
cperciva: I'd call this something like `uint8_t dummybuf[1];` to indicate that it's (a) a buffer and (b) not intended to actually be used.  It's far from clear what a variable `uint8_t shutdown` should be used for.

tests/nc-client/main.c
```diff
@@ -16,11 +18,40 @@ struct senddata {
...
+	/* We are no longer reading. */
+	send->read_cookie = NULL;
+
+	/* Check results. */
+	if (lenread == -1) {
+		warnp("network_read received");
+		goto err0;
+	} else {
+		/* We should have received nothing. */
+		assert(lenread == 0);
```
cperciva: This shouldn't be an assertion: We're asking network_read to read a byte of data, so if it returns 1 it doesn't indicate that there is a bug *internal to this program*.  (However, it indicates that the server is misbehaving, since the server shouldn't send any data back; so we should treat a nonzero value as an error.)

tests/nc-client/main.c
```diff
@@ -38,14 +69,26 @@ send_input(void * cookie)
...
 			goto err0;
 		}
+
+		/* We're not going to send anything else. */
+		if (shutdown(send->socket, SHUT_WR)) {
+			warnp("shutdown");
+			goto err0;
+		}
+
+		/* Attempt to read from socket. */
```
cperciva: I'd add a bit more detail here, along the lines of "the server should not send any data back, but attempting to read will detect when the other end of the socket is closed".

gperciva:
updated with fixes, but will need rebase.

tests/nc-client/main.c
```diff
@@ -38,7 +38,10 @@ static int callback_stopping(void * cookie, ssize_t lenread)
 		goto err0;
 	} else {
 		/* We should have received nothing. */
-		assert(lenread == 0);
+		if (lenread != 0) {
+			warnp("network_read received non-zero data");
```
cperciva: `warn0`.  No reason to think that errno has a meaningful value here.

cperciva:
Needs a minor fix and history tidying, but otherwise looks good.

gperciva:
Updated PR

PR #117, opened 2017-03-05
libcperciva getopt: silence scan-build warnings

gperciva:
Previously, scan-build in clang would produce warnings in the form:

```
main.c:159:17: warning: Null pointer passed as an argument to a 'nonnull' parameter
                        if ((opt_p = strdup(optarg)) == NULL)
```

This is not a particularly deterministic bug; creating the obvious minimal
example fails to reproduce this warning.  It seems to require some additional
code to trigger this warning, and a minimal reproduction in one version of
clang can fail to show this problem in a later version of clang.

### DIFF
--- a/libcperciva/util/getopt.h
+++ b/libcperciva/util/getopt.h
@@ -1,6 +1,7 @@
 #ifndef _GETOPT_H_
 #define _GETOPT_H_
 
+#include <assert.h>
 #include <setjmp.h>
 #include <stddef.h>
 
@@ -107,8 +108,10 @@ extern int optind, opterr, optreset;
 #define _GETOPT_OPTARG(os, ln)	__GETOPT_OPTARG(os, ln)
 #define __GETOPT_OPTARG(os, ln)						\
 	case ln:							\
-		if (getopt_initialized)					\
+		if (getopt_initialized) {				\
+			assert(optarg != NULL);				\
 			goto getopt_skip_ ## ln;			\
+		}							\
 		getopt_register_opt(os, ln - getopt_ln_min, 1);		\
 		DO_LONGJMP;						\
 	getopt_skip_ ## ln


gperciva:
Here's a fairly small (yet still readable) example, which compiles without complaint with clang 3.8.0 with:

    $ clang -Weverything -Werror -Wno-#warnings -Wno-gnu-label-as-value -c small.c 

but produces the error with scan-build (same version)

```
#include <stdlib.h>
#include <string.h>
#include <unistd.h>

#include "getopt.h"

int
main(int argc, char * argv[])
{
	size_t opt_p = 0;
	const char * opt_k = NULL;
	const char * ch;

	while ((ch = GETOPT(argc, argv)) != NULL) {
		GETOPT_SWITCH(ch) {
		GETOPT_OPTARG("-k"):
			if (opt_k)
				exit(1);
			opt_k = optarg;
			break;
		GETOPT_OPTARG("-p"):
			opt_p = strlen(optarg);
			(void)opt_p;
			break;
		GETOPT_DEFAULT:
			break;
		}
	}
}
```

gperciva:
Here's a smaller example (from creduce) which is less understandable.  I'm intrigued (in a slightly horrified way) about what the first line is doing.

```
# 1 "" 3
libcperciva_optarg, main_opt_k;
strlen(*) __attribute__ ((__nonnull__ ));
main(int argc, char * argv[]) {
    char ch;
    while((ch=libcperciva_getopt(argc,argv))!=((void*)0)) {
        if (main_opt_k)
		main_opt_k = libcperciva_optarg;
        strlen(libcperciva_optarg);
    }
}
```

cperciva:
The first line of that reduced test case is saying "these things exist and have type int", since "int" is the default type if no type is specified. :-)

I like the fix but not the commit message; you haven't explained why the assertion is guaranteed to be valid.  Something along the lines of "If a command-line option is specified (via GETOPT_OPTARG) to take an argument and no argument is available, getopt will return opt_missing and GETOPT_SWITCH will land us in either GETOPT_MISSING_ARG or GETOPT_DEFAULT; so we are guaranteed to have a command-line argument if we land in GETOPT_OPTARG.  Asserting this helps to prevent static-analysis software from getting confused."

gperciva:
I know that unspecified-type variables are int.  The first line -- which is still a **necessary** line -- in the reduced test case is:

```
# 1 "" 3
```

gperciva:
Updated PR with better commit message.

cperciva:
Oh, you meant *that* first line.  I thought you were counting starting at zero.  Judging by clang's behaviour when I remove it, I'm guessing it's an internal directive telling clang to disable a bunch of warnings?

PR #115, opened 2017-02-28
use Travis-CI strict compile warnings/errors

gperciva:

PR #118, opened 2017-03-14
libcperciva poll: explicit cast to (nfds_t)nfds

gperciva:
On MacOS X, nfds_t is (unsigned int) rather than (unsigned long).

This cast is safe because the only place in which we increase nfds is in
growpollfd(), and we do so in response to adding a new (int)fd to the pollfd
array, and thus therefore nfds won't exceed INT_MAX.


gperciva:
Split from #115

PR #119, opened 2017-03-24
Misc

gperciva:


gperciva:
I think these are sufficiently small and well-contained that they don't need separate PRs, but of course I could do that if you want.

PR #121, opened 2017-03-26
Daemonize cleanup

gperciva:


gperciva:
Based on offline communication, you're going to say "ick, no.  Throw that mess away!".  But since I already had it written, I figured I'd show it to you for the 1% chance that you might like any aspects of it.  :)

cperciva:
Nope, don't like it. :-)

PR #112, opened 2017-02-17
Remove limitation on the maximum number of concurrent connections handled by spiped

cperciva:

PR #123, opened 2017-04-12
Parsenum

gperciva:


gperciva:
Once this is merged, I'll update #112 on top of it.

cperciva:
Going to merge this as soon as the build finishes, but we should add opt_o_set and opt_r_set in order to catch e.g. 'spiped -o 0 -o 0'.

cperciva:
(And by "we" I mean "you", and by "add" I mean "in a separate PR".)

PR #124, opened 2017-04-14
Don't allow multiple -o or -r options

gperciva:
This does not alter the -n option, as that will happen in a separate PR
(wherein -n 0 will gain semantic meaning).


gperciva:
This is another prereq for #112.

PR #129, opened 2017-04-20
Makefile discourage git

gperciva:


gperciva:
Not certain if you'll like this.  It'll take care of people stumbling across github's "download zip" feature (or simply "git clone") and unknowingly ending up with an unstable version of spiped.  It's easy enough for us to add `export SPIPED_DEVEL=1` to `~/.profile`, so I think there's no harm in it... but I must admit that I'm in the "60% for, 40% against" category.

gperciva:
Oh, it'll also work on the github "release" tarballs, as those come from git.  At least, it'll work on all future releases.

TimWolla:
> Oh, it'll also work on the github "release" tarballs, as those come from git.

I'm fairly sure that GitHub uses the `git-archive(1)` command to generate those exports. You might want to take a look at this, maybe it works:

> ATTRIBUTES
>       export-ignore
>           Files and directories with the attribute export-ignore won’t be added to archive files. See gitattributes(5) for details.

cperciva:
It's good to see what this looks like, but I don't like it.  For one thing, this won't solve the problem for old releases (and for future releases there isn't much of a problem anyway); and for another thing, I hate the idea of adding code to our tree just to work around a usability bug in github.

PR #132, opened 2017-04-21
build: add tests/ to tarball

gperciva:


gperciva:
I checked that the resulting tarball compiles and runs `make test`.

PR #139, opened 2017-06-06
Run performance tests

gperciva:


gperciva:
I made this a PR so that it's easier for @cperciva to take a look at it.  Data coming soon.

gperciva:
Looks like FreeBSD is choking on buffer sizes above 2048 ?  Testing on Amazon Linux vs. FreeBSD 11. on EC2 `m4.large`.

### Summary
Timing is pretty consistent:
gnuplot: `plot 'linux-1.txt' u 1:4 w lp lt 1, 'linux-2.txt' u 1:4 w lp lt 1, 'freebsd-1.txt' u 1:4 w lp lt 2, 'freebsd-2.txt' u 1:4 w lp lt 2`
linux-freebsd

### Raw data (other than making tabs line up)
Linux run 1

```
Sending 68719 Mb via send-zeros -> recv-zeros
block	count		time (s)	speed (Mb/s)
512	134217728	77.540000	886.245509
1024	67108864	39.810000	1726.186303
2048	33554432	23.000000	2987.803336
4096	16777216	16.490000	4167.342434
8192	8388608		11.820000	5813.830519
```

Linux run 2
```
Sending 68719 Mb via send-zeros -> recv-zeros
block	count		time (s)	speed (Mb/s)
512	134217728	77.380000	888.078014
1024	67108864	39.160000	1754.838527
2048	33554432	23.080000	2977.446999
4096	16777216	16.450000	4177.475789
8192	8388608		11.790000	5828.623981
```

FreeBSD run 1
```
Sending 68719 Mb via send-zeros -> recv-zeros
block	count		time (s)	speed (Mb/s)
512	134217728	120.000000	572.662306
1024	67108864	52.492188	1309.137226
2048	33554432	28.085938	2446.757447
4096	16777216	56.164062	1223.548897
8192	8388608		51.140625	1343.735567
```

FreeBSD run 2
```
Sending 68719 Mb via send-zeros -> recv-zeros
block	count		time (s)	speed (Mb/s)
512	134217728	119.585938	574.645131
1024	67108864	52.226562	1315.795516
2048	33554432	28.390625	2420.498905
4096	16777216	55.562500	1236.795982
8192	8388608		50.250000	1367.551776
```

cperciva:
Siva Mahadevan will be looking into what's happening in the FreeBSD kernel to create this behaviour... to quote Ed Maste, "a good place to explore using DTrace to investigate system performance".

cperciva:
Weirdly, neither I nor anyone else at BSDCam are now able to reproduce this FreeBSD performance anomaly.

cperciva:
... never mind, I can now reproduce this in EC2 running 11.1-RELEASE.

gperciva:
... huh, do you think it might be a virtualization thing?  I mean, did nobody manage to reproduce it with bare-metal 11.1-RELEASE?

cperciva:
I think there's two issues here:

1. FreeBSD's unix socket buffers are fixed at 8 kB (for some definition of "fixed": You can write more than this into a socket, but then it blocks until the buffers empty), which results in a transition when we move from small writes (where the reader and writer can both be running at the same time) to larger writes (where the processes ping-pong).

2. With large writes, I'm seeing a total CPU usage of around 80-90%, which suggests that the ping-ponging is either very slow or is occasionally getting stuck; in the past I've seen things which look like inter-cpu wakeups going missing under Xen, so this may be related.

cperciva:
BTW the first issue can be fixed by setting `sysctl net.local.stream.sendspace=65536` and similarly for `recvspace`.

gperciva:
Finally moving forward on this one!  I'll put the "visualize results of performance tests" in a separate PR (later), because I'm not certain what you'll feel about python.

In `01-spiped-throughput.sh`, we compare the speed of direct socket connections vs. spiped connections.
- we repeat each test 3 times, since that seemed like a decent value to give us an idea of how variable the tests are
- we test 7 block sizes, from 1024 to 65536 inclusive.
- we send 16 MB each time, because that ended up being 4 minutes when run in freebsd inside virtualbox on my desktop.

The last point is obviously the most arbitrary.  I recall that early on, you suggested sending 1 Gb, which would make the tests slower but perhaps more meaningful.  It's easily tweakable, either way.

perftests/01-spiped-throughput.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/sh
+
+### Constants
+total=$( echo "2^24" | bc )
```
gperciva: I'm going to tweak these so that the user specifies which type of system, and it sets a sensible value accordingly.

gperciva:
I'll bring this back in a few months.

PR #140, opened 2017-06-07
Network mpool

gperciva:


libcperciva/network/network_read.c
```diff
@@ -21,6 +22,8 @@ struct network_read_cookie {
 	size_t bufpos;
 };
 
+MPOOL(network_read_cookie, struct network_read_cookie, 16);
```
gperciva: The `MPOOL(eventrec...)` line uses 4096 for the cache size.  I'm not certain what the trade-off is for smaller sizes; I went with 16 for `network_*_cooke` as a random guess.

PR #144, opened 2017-08-07
Makefiles depth

gperciva:


gperciva:
When running

    make Makefiles

some source files need to find cpusupport-config.h.  For the "Makefiles" target, we create a fake file in the top-level directory.

Previously, it was hard-coded that the top-level directory was

    ../

which worked fine for spipe/ and spiped/.  By chance, none of the programs under

    tests/*/

needed cpusupport-config.h, so this wasn't an issue.

However, we now need the flexibility to refer to the top-level directory (in other projects).  I'm putting it in spiped first, since it's the most mature project we have which uses Makefile.BSD to produce POSIX-compliant Makefiles.

gperciva:
hmm, something's weird -- it definitely works in spiped, and I _thought_ it was working in libcperciva, but now it's not.  I'll take another look at it with fresh eyes tomorrow.  (if you're energetic from jetlag and think this looks perfectly fine, go ahead and pull it, but I'm ok with this being on the backburner while I investigate further)

gperciva:
Found the problem -- I deleted (the new) `tests/Makefile.inc` as part of a "clean up and test" attempt, so of course the build system didn't know how to build `source-*` and `cflags-*`.

So this PR should be good to merge.

gperciva:
PR revised according to offline comments (i.e. tracking `SUBDIR_DEPTH` in `Makefile.inc`, instead of using `sed`).

gperciva:
I added the `${LIBCPERCIVA_DIR}` simplification (for maintainability; has no effect on any `Makefile`).

spipe/Makefile.BSD
```diff
@@ -10,6 +10,9 @@ LDADD	=	-lrt
 LDADD_REQ	=	-lcrypto
 LDADD_REQ	+=	-lpthread
 
+# Useful relative directories
+LIBCPERCIVA_DIR	=	../libcperciva
```
cperciva: Wasn't the idea that this was going to make use of SUBDIR_DEPTH?

gperciva: Unfortunately, `.include <bsd.prog.mk>` comes at the end, so SUBDIR_DEPTH isn't defined when make parses this line.  Also, it wants to find `SRCS += foo.c` before it does the late parsing of include files.

I'm not saying it's **impossible** to get it to work, but after 3 or 4 hours, it was looking like it would involve a lot of black magick and I'd need to dig into the source file(s) of BSD make.  And since it could be achieved with a single manually-written line in each `Makefile.BSD`, I figured that was the best option.

(Note that `SUBDIR_DEPTH` in each subdir (via `Makefile.inc`) is still useful, as that allows the top-level `Makefile.BSD` to include the correct directory to find the top-level `cpusupport-config.h`.)

gperciva:
Updated PR to include `LIBCPERCIVA_DIR` in the main Makefile.

There's 3 things this PR does:
- allows subdirectory-programs to include the directory containing `cpusupport-config.h.`  (currently if any `test/` program needs it, it'll fail)
- allows more shared code between subdirectory-programs (maintainability, particularly when adding a new `test/` program)
- allows more shared code between projects (i.e. fewer diffs between `spiped/Makefile.BSD` and `libcperciva/Makefile.BSD`).

PR #150, opened 2017-10-30
Test dnsthread resolve

gperciva:


gperciva:
I *think* I wrote this so that I could convince myself that it wasn't leaking memory (after the clang static analyzer flagged a potential leak).  To be honest, I can't remember entirely; I found the branch while I was cleaning up other stuff in spiped, and it was 90% done, so I figured I'd finish it.

Note that this tests `dnsthread_resolve()`, which is not used elsewhere in the spiped git tree.  If you think it's not worth keeping that around (because `spiped/dispatch.c` only uses `dnsthread_resolveone()`), then I'm quite content to rewrite the test (and potentially also delete `dnsthread_resolve()`).

tests/07-dnsthread.sh
```diff
@@ -0,0 +1,21 @@
...
+### Constants
+addr_output="${s_basename}-addrs.txt"
+
+
+### Actual command
+scenario_cmd() {
+	# Check address
+	setup_check_variables
+	(
+		${dnsthread_resolve} localhost:80 > ${addr_output}
```
cperciva: Maybe this should test a variety of lookups?  Raw IP addresses, names which need to actually get looked up...?

gperciva: I've added some more test cases, but I avoided names that actually need to get looked up -- that would require an active internet connection, and (so far) none of the tests require one.

That said, I suppose that I could support it with the same general idea as the setuidgid test -- if it fails due to the relevant errno, I could output a special code to indicate "skip this test".  I'd rather have that as a separate `tests/??-*.sh` filename, though.  Like, `07-dnsthread-local.sh` and `08-dnsthread-network.sh`.

gperciva:
I wish github didn't hide comments on outdated patches.  PR updated, needs rebase.

I've added some more test cases, but I avoided names that actually need to get looked up -- that would require an active internet connection, and (so far) none of the tests require one.

That said, I suppose that I could support it with the same general idea as the setuidgid test -- if it fails due to the relevant errno, I could output a special code to indicate "skip this test". I'd rather have that as a separate tests/??-*.sh filename, though. For example, 07-dnsthread-local.sh and 08-dnsthread-network.sh.

tests/07-dnsthread-local.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+# Goal of this test:
+# - ensure that dnsquery can find simple addresses
+
+### Constants
+addr_output="${s_basename}-addrs.txt"
+addrs="	localhost:80
+	1.2.3.4:80
```
cperciva: Not valid, IP addresses need to be in square brackets.

cperciva:
Can we have an environment variable, something like `TEST_DNSTHREAD_ENABLEREMOTE` and select whether to run the full set of tests or just the local tests based on that?  Leave it turned off by default, but I think we can turn it on in Travis CI?

gperciva:
Done.  Shall I rebase the history now?

.travis.yml
```diff
@@ -6,12 +6,15 @@ matrix:
...
       env: CFLAGS="-std=c99 -O2 -Wall -Wextra -Werror -Weverything -Wno-#warnings -Wno-pedantic -Wno-padded -Wno-format-nonliteral -Wno-disabled-macro-expansion -Wno-missing-noreturn -Wno-reserved-id-macro -Wno-cast-align"
+           TEST_DNSTHREAD_ENABLE_REMOTE=1
     - os: linux
       compiler: gcc
       env: CFLAGS="-std=c99 -O2 -Wall -Wextra -Werror -Wpedantic -pedantic-errors -Wno-clobbered"
+           TEST_DNSTHREAD_ENABLE_REMOTE=1
     - os: osx
       compiler: clang
       env: CFLAGS="-std=c99 -O2 -Wall -Wextra -Werror -Weverything -Wno-#warnings -Wno-pedantic -Wno-padded -Wno-format-nonliteral -Wno-disabled-macro-expansion -Wno-missing-noreturn -Wno-reserved-id-macro"
+           TEST_DNSTHREAD_ENABLE_REMOTE=1
```
cperciva: Can't this be a global environment variable rather than being listed three times?

cperciva:
Looks plausible.  Do we want to check that DNS resolution actually produces the correct results rather than merely *thinking* that it succeeded?  I could create fake names under tarsnap.net which are guaranteed to resolve to particular IPs.

gperciva:
I don't think we need to go as far as checking actual IPs -- if the code actually starts producing garbage, I'm sure that we'll find out really quickly.  :)

.travis.yml
gperciva: done

gperciva:
updated PR

PR #152, opened 2017-12-08
Add -U username[:groupname]

gperciva:


gperciva:
This is an example of https://github.com/Tarsnap/libcperciva/pull/78 in use.  We shouldn't pull it until that one is merged.

PR #153, opened 2017-12-11
Avoid SIGSEGV on shutdown with target re-resolution

michaelweiser:
spiped will segfault on shutdown if target address re-resolution is in
use. This is due to two cases of accessing memory that will be freed by
other parts of the code on the first re-resolution cycle. To avoid this:

The DNS timer has to be marked as expired so it's not freed on shutdown
if already expired. Incidentally, the new instance's is recorded so it
can be cancelled on shutdown if not already expired.

Also, the initial target address list as taken from the command line has
to be marked as under dispatch's purview after dispatch_accept() has
been called.

### DIFF
--- a/spiped/dispatch.c
+++ b/spiped/dispatch.c
@@ -61,8 +61,8 @@ callback_resolve(void * cookie, struct sock_addr ** sas)
 	}
 
 	/* Wait a while before resolving again. */
-	if (events_timer_register_double(callback_resolveagain,
-	    A, A->rtime) == NULL)
+	if ((A->dnstimer_cookie = events_timer_register_double(
+	    callback_resolveagain, A, A->rtime)) == NULL)
 		goto err0;
 
 	/* Success! */
@@ -78,6 +78,9 @@ static int
 callback_resolveagain(void * cookie)
 {
 	struct accept_state * A = cookie;
+
+	/* This timer is expired. */
+	A->dnstimer_cookie = NULL;
 
 	/* Re-resolve the target address. */
 	return (dnsthread_resolveone(A->T, A->tgt, callback_resolve, A));
@@ -308,6 +311,7 @@ dispatch_shutdown(void * dispatch_cookie)
 		events_timer_cancel(A->dnstimer_cookie);
 	if (A->T != NULL)
 		dnsthread_kill(A->T);
+	sock_addr_freelist(A->sas);
 	free(A);
 }
 
--- a/spiped/main.c
+++ b/spiped/main.c
@@ -315,6 +315,9 @@ main(int argc, char * argv[])
 		goto err5;
 	}
 
+	/* dispatch is now maintaining sas_t. */
+	sas_t = NULL;
+
 	/* Register a handler for SIGTERM. */
 	if (graceful_shutdown_initialize(&callback_graceful_shutdown,
 	    dispatch_cookie)) {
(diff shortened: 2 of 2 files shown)


michaelweiser:
Travis build failure on macOS is unrelated. Fix here: https://github.com/Tarsnap/spiped/pull/154.

gperciva:
Thanks, @michaelweiser.  Could you please rebase onto current master and do a force push to trigger a new Travis-CI build?

spiped/dispatch.c
```diff
@@ -61,8 +61,9 @@ callback_resolve(void * cookie, struct sock_addr ** sas)
 	}
 
 	/* Wait a while before resolving again. */
-	if (events_timer_register_double(callback_resolveagain,
-	    A, A->rtime) == NULL)
+	A->dnstimer_cookie = events_timer_register_double(callback_resolveagain,
+	    A, A->rtime);
+	if (A->dnstimer_cookie == NULL)
```
gperciva: Our code style would write this as:
```
if ((A->dnstimer_cookie = events_blah(blah)) == NULL)
```
I personally prefer your version, but I know that it won't be accepted as-is.  Also, having a consistent style is more important than any one particular stylistic preference.  :)

spiped/dispatch.c
```diff
@@ -79,6 +80,10 @@ callback_resolveagain(void * cookie)
 {
 	struct accept_state * A = cookie;
 
+	/* this timer is expired for sure  - mark it as such so we don't try to
```
gperciva: For a multiline comment, please leave the `/*` and `*/` on separate lines:
```
/*
 * This timer is expired - mark it as such so
 * we don't try to cancel it on shutdown.
 */
```

gperciva:
Hi @michaelweiser!  I made some sample changes (the style of comments), and uploaded to https://github.com/Tarsnap/spiped/tree/michaelweiser-master.  If you like them, please rebase onto your patch (and the whole thing on to current master).

michaelweiser:
Hi @gperciva, all done as you suggested. I squashed your style changes into my commit for atomicity and rebased onto current master. Travis is happy now, too.

gperciva:
Thanks, LGTM!  I know that @cperciva is busy preparing a paper for a conference, so no guarantees as to when he can take a look at it, but I'll make sure it's in the first batch.

michaelweiser:
No prob whatsover: Over debugging I realized I don't even need the re-resolution feature since my spiped only ever connects to localhost and just disabled it. Just wanted to share my findings so the next poor soul doesn't run into it. :)

spiped/main.c
```diff
@@ -341,7 +344,8 @@ main(int argc, char * argv[])
 	free(K);
 
 	/* Free arrays of resolved addresses. */
-	sock_addr_freelist(sas_t);
+	if (sas_t != NULL)
```
gperciva: I just noticed that `sock_addr_freelist()` does nothing if the pointer is already NULL, so this line isn't strictly necessary.

spiped/main.c
```diff
@@ -357,7 +361,8 @@ main(int argc, char * argv[])
 err4:
 	free(K);
 err3:
-	sock_addr_freelist(sas_t);
+	if (sas_t != NULL)
```
gperciva: I just noticed that `sock_addr_freelist()` does nothing if the pointer is already NULL, so this line isn't strictly necessary.

michaelweiser:
Right on!

cperciva:
Seems like the `sock_addr_freelist(sas_t);` in main.c could just be unconditionally removed now?  Is there any way for `sas_t` to *not* be NULL when we get there?

PR #154, opened 2017-12-12
Fix timespec/timeval nsec/usec conversion

michaelweiser:
Explicitly cast conversion result of nsec / 1000 to suseconds_t to
account for possible difference in variable width and signedness as
observed on macOS causing warnings such as:

../libcperciva/util/monoclock.c:36:28: error: implicit conversion loses
integer precision: 'long' to '__darwin_suseconds_t' (aka 'int')
      [-Werror,-Wshorten-64-to-32]
                tv->tv_usec = tp.tv_nsec / 1000;
                            ~ ~~~~~~~~~~~^~~~~~

### DIFF
--- a/libcperciva/util/monoclock.c
+++ b/libcperciva/util/monoclock.c
@@ -33,7 +33,7 @@ monoclock_get(struct timeval * tv)
 #ifdef USE_MONOTONIC
 	if (clock_gettime(CLOCK_MONOTONIC, &tp) == 0) {
 		tv->tv_sec = tp.tv_sec;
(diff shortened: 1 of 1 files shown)


gperciva:
Thanks!  I have this fix waiting in https://github.com/Tarsnap/libcperciva/pull/89, after which I'll put it in spiped.

The current build failure is also unrelated; Travis-CI no longer contains gcc-4.9 (which was hardcoded in our YAML a long time ago).  I'm tracking down whether they offer gcc at all, or whether it's just a front-end to llvm at the moment.

gperciva:
Thanks, this is now fixed in master.

Issue #160, opened 2018-02-19
Add CPPFLAGS / LDFLAGS placeholders to Makefile


TimWolla:
copying my request from IRC:

> &lt;TimWolla&gt; cperciva_, I wanted to compile Tarsnap's Debian based docker image using the recommended Debian hardening flags (https://wiki.debian.org/Hardening#dpkg-buildflags) and learned about this mailing list thread: http://mail.tarsnap.com/spiped/msg00044.html. Could you possibly revisit adding LDFLAGS as well as CPPFLAGS to your Makefiles?

You can find the patches the Debian maintainer does here: https://gitlab.com/spiped/spiped/blob/debian/debian/patches/honor-compiler-flags.patch. It's obviously not applicable to your autogenerated Makefiles, but it should give an idea what is required by the Debian project.

TimWolla:
We also had a short discussion about adding (a few) of those flags by default. For reference, this is what Ubuntu Xenial (my workstation) spits out for the various variables:

```
[timwolla@/s/D/o/docker-spiped (master)]dpkg-buildflags                 23:46:23
CFLAGS=-g -O2 -fstack-protector-strong -Wformat -Werror=format-security
CPPFLAGS=-Wdate-time -D_FORTIFY_SOURCE=2
CXXFLAGS=-g -O2 -fstack-protector-strong -Wformat -Werror=format-security
FCFLAGS=-g -O2 -fstack-protector-strong
FFLAGS=-g -O2 -fstack-protector-strong
GCJFLAGS=-g -O2 -fstack-protector-strong
LDFLAGS=-Wl,-Bsymbolic-functions -Wl,-z,relro
OBJCFLAGS=-g -O2 -fstack-protector-strong -Wformat -Werror=format-security
OBJCXXFLAGS=-g -O2 -fstack-protector-strong -Wformat -Werror=format-security
```

This is what a Debian Stretch Docker container gives:

```
root@ac8e04e96d86:/pwd# dpkg-buildflags
CFLAGS=-g -O2 -fdebug-prefix-map=/pwd=. -fstack-protector-strong -Wformat -Werror=format-security
CPPFLAGS=-Wdate-time -D_FORTIFY_SOURCE=2
CXXFLAGS=-g -O2 -fdebug-prefix-map=/pwd=. -fstack-protector-strong -Wformat -Werror=format-security
FCFLAGS=-g -O2 -fdebug-prefix-map=/pwd=. -fstack-protector-strong
[...]

PR #161, opened 2018-02-20
Build: add CPPFLAGS, LDFLAGS

gperciva:


gperciva:
Inspired by #160 (@TimWolla) and Peter Pentchev's (the Debian spiped maintainer) patches.

I won't claim to have an opinion on whether `CPPFLAGS` or `CFLAGS` should be the final argument; I put `CFLAGS` at the end because that's what Peter did.

PR #162, opened 2018-04-10
Readme examples

gperciva:


gperciva:
Now that we have nice diagrams on the spiped website, let's point people to them.

PR #164, opened 2018-04-23
Libcperciva import

gperciva:


gperciva:
This includes the mpool linked list and autotuning updates.  Do you want them included in this import, or should that happen in a separate PR?

cperciva:
Unless there's a specific reason to do otherwise (e.g., code currently broken and being worked on) I'd say that merges from libcperciva should always update to the most recent version of the subset which is included.

PR #165, opened 2018-04-24
Setuidgid

gperciva:


spiped/main.c
```diff
@@ -28,6 +29,7 @@ usage(void)
 	    "    [-DFj] [-f | -g] [-n <max # connections>] "
 	    "[-o <connection timeout>]\n"
 	    "    [-p <pidfile>] [-r <rtime> | -R]\n"
+	    "        [rU {<username> | <:groupname> | <username:groupname>}]\n"
```
cperciva: `rU`?  Presumably that should be `-U`?

spiped/main.c
```diff
@@ -28,6 +29,7 @@ usage(void)
 	    "    [-DFj] [-f | -g] [-n <max # connections>] "
 	    "[-o <connection timeout>]\n"
 	    "    [-p <pidfile>] [-r <rtime> | -R]\n"
+	    "    [-U {<username> | <:groupname> | <username:groupname>}]\n"
```
gperciva: Do you prefer `-u` or `-U` for setuidgid?

cperciva: Good question.  I've asked twitter which they would expect, and so far `-u` is leading over `-U` by a 10:1 ratio.

cperciva: (Let's wait until the poll closes in a few hours in case this is just a matter of the early numbers being wildly off though.)

cperciva: The poll is 83% in favour of calling the option `-u`.

One thing a few people mentioned though is that we should accept numeric IDs as well.  I guess that's something to do into a revised setuidgid...

gperciva: No to numeric ids.  `"0"` is a perfectly valid **username** in POSIX.

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_437:

> 3.437 User Name
> A string that is used to identify a user; see also User Database. To be portable across systems conforming to POSIX.1-2017, the value is composed of characters from the portable filename character set. The <hyphen-minus> character should not be used as the first character of a portable user name.

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282
> 3.282 Portable Filename Character Set
> The set of characters from which portable filenames are constructed.
>
> A B C D E F G H I J K L M N O P Q R S T U V W X Y Z
> a b c d e f g h i j k l m n o p q r s t u v w x y z
> 0 1 2 3 4 5 6 7 8 9 . _ -
>
> The last three characters are the <period>, <underscore>, and <hyphen-minus> characters, respectively. See also Pathname.

For fun:
```
$ sudo adduser --force-badname 0
Allowing the use of a questionable username.
Adding user `0' ...
Adding new group `0' (1003) ...
Adding new user `0' (1003) with group `0' ...
Creating home directory `/home/0' ...
Copying files from `/etc/skel' ...
Enter new UNIX password: 
Retype new UNIX password: 
passwd: password updated successfully
Changing the user information for 0
Enter the new value, or press ENTER for the default
	Full Name []: 
	Room Number []: 
	Work Phone []: 
	Home Phone []: 
	Other []: 
Is the information correct? [Y/n] y
td@mac: ~
$ su 0
Password: 
0@mac:/home/td
```

gperciva: Apparently Ubuntu can be cajoled to going beyond POSIX -- from https://unix.stackexchange.com/a/287086 :

```
root@c2:~# useradd 😀
...
$ whoami
😀
```

and
```
root@c2:~# useradd '&#%^()!@~*?<>=|'
...
$ whoami
&#%^()!@~*?<>=|
```

I'm not suggesting we go that far, of course!  But POSIX clearly allows digit-only **username**.

Previous discussion, including you saying "don't be systemd" :)
https://github.com/Tarsnap/libcperciva/pull/78#discussion_r143388384

cperciva: I said "don't be systemd", but I meant to not misparse user names as if they're user IDs...

POSIX says that `chown` will try to parse a string as a numeric ID only if no user exists with that name -- it seems like we should probably behave the same way.

PR #168, opened 2018-05-01
Build unify

gperciva:


gperciva:
This mainly imports changes from libcperciva.  The only non-imported changes are defining & using `${TEST_CMD}`, and adding the `### Shared code between Tarsnap projects.` markers.

PR #175, opened 2018-07-10
Valgrind prep

gperciva:


gperciva:
Eliminates a false positive on FreeBSD with `pthread_join()`.

PR #179, opened 2018-10-15
spiped: better error for strdup

gperciva:


gperciva:
This replaces `OPT_EPARSE` with `warnp("strdup()");`.

I'm not trying to be lazy here; it's just that getting the text version of ENOMEM would be a better error message than "Error parsing argument".

It looks like I should be making the same change everywhere in kivaloo.  Unless you had a brilliant sneaky reason for `OPT_EPRASE` instead of `warnp` in the first place...?

cperciva:
1. OPT_EPARSE should print "spiped: Error parsing argument: -p foobar: Cannot allocate memory", so it's not like the fact of ENOMEM gets lost.

2. OPT_EPARSE contains an `exit(1)` which you seem to have lost...

gperciva:
Improved `OPT_EPARSE`, and using that for the `PARSENUM()` errors as well.

PR #183, opened 2018-12-08
Tests unix sockets

gperciva:


gperciva:
This is prep for the perftests.

This includes PR #182, so you might want to merge that one first and then I'll rebase.  I don't mind if you want to skip that one and merge this directly, though.

PR #184, opened 2018-12-08
spiped: remove the unix domain socket we created

gperciva:


gperciva:
This PR is the starting place of a discussion, not a merge candidate.

At the moment, libcperciva doesn't clean up any unix domain sockets it creates.  That triggers my OCD, but more importantly, it means that you can't run repeated tests (unless the test suite removes the domain socket files).

One option, as done in this PR, is for the calling code (in this case, `main.c`) to delete the file.

I think it would be "cleaner" to handle this at the sock layer, though.  Either with a

    void sock_remove(const char * sockname);

which is basically just the above code, or automatically.  I *did* manage to make it automatic, but it seemed a bit messy:
- to `struct sock_addr` in `libcperciva/util/sock_internal.h`, add a `int created_unix`.
- to `int sock_listener(const struct sock_addr * sa)` in `libcperciva/util/sock.c`, remove the `const`.
- in the same function, add a `created_unix = 1` after the `bind()`.  Err, inside an appropriate `if` statement.
- in `sock_addr_free()`, do the `unlink()` if appropriate.

Which solution do you prefer?

gperciva:
PS: actually, this is still a bit theoretical: in the test suite, we can create the unix domain sockets inside the test output directory, which we remove and recreate every time we run the tests.  So it's not an actual problem for the perftests.

gperciva:
In offline discussion, @cperciva said not to do this inside spiped itself; the calling shell script should do it.

PR #187, opened 2018-12-25
Perftests binaries

gperciva:


gperciva:
These are extracted (and slightly updated) from #139.  I thought it would be easier to review the binaries separately from the shell script.  Sample usage:

    recv-zeros /tmp/sock 8192 &
    sleep 1
    send-zeros /tmp/sock 8192 1024 >> times.txt

which produces the output

    8192    1024    0.0077  1086.04

which stands for: block size, number of blocks, time (seconds), speed (MB/s).

perftests/recv-zeros/main.c
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	/* Accept connections. */
+	if ((socket_recv = accept(socket, NULL, NULL)) == -1) {
+		warnp("accept");
+		goto err3;
+	}
+
+	/* Receive data, but do nothing with it. */
+	r = buflen;
```
cperciva: I would write this

```
do {
    r = recv(...);
} while (r == buflen);
```

perftests/recv-zeros/Makefile.BSD
```diff
@@ -0,0 +1,39 @@
...
+SRCS	=	main.c
+
+# Data structures
+.PATH.c	:	../../libcperciva/datastruct
+SRCS	+=	elasticarray.c
+SRCS	+=	ptrheap.c
+SRCS	+=	timerqueue.c
+IDIRS	+=	-I../../libcperciva/datastruct
+
+# Event loop
```
cperciva: Are we using any of this libcperciva library code?

gperciva:
We could also use network sockets instead, and/or slap a pair of `spiped` daemons between the send and receive.

The binary doesn't print any column headers because it's intended to be called multiple times with different block sizes and different unix/network sockets and direct/spiped connections.

perftests/recv-zeros/Makefile.BSD
gperciva: whoops, that was left over from a much earlier version.

perftests/send-zeros/main.c
```diff
@@ -0,0 +1,139 @@
...
+	while (to_send > 0) {
+		if (write(socket, buffer, buflen) != (ssize_t)buflen) {
+			warnp("write failed");
+			goto err3;
+		}
+		to_send--;
+	}
+
+	/* Get ending time. */
+	if (monoclock_get(&end)) {
```
cperciva: This measures how long it takes to *send* the data, but not how long it takes for the data to be received -- it might be sitting in kernel buffers.

I think we discussed having a `shutdown(..., SHUT_WR)` and then calling `read` which will return with 0 (aka EOF) when the remote end is closed?

perftests/recv-zeros/Makefile.BSD
gperciva: same in send-zeros, sorry.

perftests/send-zeros/main.c
gperciva: looks like we implemented that in nc-client, but I didn't remember to adjust send-zeros accordingly.

gperciva:
I updated the `Makefile.BSD` and `Makefile` files before I remembered that I usually leave fixes as "REBASE" commits.

cperciva:
LGTM, please clean up the history.

PR #195, opened 2019-06-04
libcperciva import

gperciva:


release-tools/metabuild.sh
```diff
@@ -6,6 +6,18 @@ CFLAGS_HARDCODED=$3
 
 OUT=Makefile
 
+# Check environment
```
cperciva: ?

gperciva:
Updated PR with correct commits (not squashed).

PR #203, opened 2019-08-11
Warn if spipe has no connection

gperciva:


gperciva:
Includes #202.

With this PR, if there's no server on `8001`, we get:
```
$ ./spipe/spipe -t 127.0.0.1:8001 -k /dev/null
spipe: Could not connect
spipe: Error running event loop
$ echo $?
1
```

In current `master`, the same command exits with no message and an exit code of 0.

proto/proto_conn.h
```diff
@@ -5,6 +5,13 @@
 struct proto_secret;
 struct sock_addr;
 
+/* Reason why the connection was dropped. */
+enum {
+	PROTO_CONN_CANCELLED = 0,	/* Client code */
+	PROTO_CONN_CONNECT_FAILED,	/* Could not connect */
+	PROTO_CONN_ERROR,		/* Unspecified reason */
+};
```
cperciva: Seems like we need a `PROTO_CONN_CLOSED` here...

proto/proto_conn.c
```diff
@@ -390,11 +390,11 @@ callback_pipestatus(void * cookie)
 
 	/* If we have an error in either direction, kill the connection. */
 	if ((C->stat_f == -1) || (C->stat_r == -1))
-		return (proto_conn_drop(C));
+		return (proto_conn_drop(C, PROTO_CONN_ERROR));
 
 	/* If both directions have been shut down, kill the connection. */
 	if ((C->stat_f == 0) && (C->stat_r == 0))
-		return (proto_conn_drop(C));
+		return (proto_conn_drop(C, PROTO_CONN_ERROR));
```
cperciva: ... which would be used here, since cleaning up after both ends of a connection have been shut down is absolutely not an error.

spipe/main.c
```diff
@@ -17,15 +17,25 @@
 #include "pushbits.h"
 
 static int
-callback_conndied(void * cookie)
+callback_conndied(void * cookie, int reason)
 {
 	int * conndone = cookie;
 
+	/* Check reason. */
+	if (reason == PROTO_CONN_CONNECT_FAILED) {
```
cperciva: This seems like it should be a `switch`.

spipe/main.c
```diff
@@ -17,15 +17,25 @@
...
+	}
+
 	/* Quit event loop. */
 	*conndone = 1;
 
 	/* Success! */
 	return (0);
+
+err0:
+	/* Failed */
```
cperciva: I wouldn't do this.  Normally when a callback returns `-1` it means "something drastic and unrecoverable happened", e.g., not being able to allocate memory for a continuation or a syscall failing which normally shouldn't.  Networks not working is not a *catastrophic* failure.

So as far as making `main` return 1, I'd go with "record that the connection could not be established and decide what value to exit from main with" rather than "throw a callback failure and watch everything abort".

As far as the UX goes, this means seeing "Could not connect" and ending with `exit(1)` but not reporting "Error running event loop".

spipe/main.c
gperciva: err, really?  This seems unnecessarily verbose:

```
        /* Check reason. */
        switch (reason) {
        case PROTO_CONN_CONNECT_FAILED:
                connection_failed = 1;
                break;
        default:
                ;
                /* FALLTHROUGH */
        }
```

cperciva: Well, it depends if you're going to do anything else with the reason codes later.  I was guessing that you probably would, since otherwise why have multiple codes which are handled identically?

gperciva: Well, the git commit message was:

Currently, most errors are the generic PROTO_CONN_ERROR, but if there's
a desire to distinguish other cases (such as timeouts vs. protocol
handshake errors), that's easily added.

but I could rephrase that to say "Currently, the only one that will receive special attention is PROTO_CONN_CONNECT_FAILED"?

cperciva: The commit message is fine.  I just thought that it would make sense to write this code in a way which makes future progress easier.

gperciva:
Updated with some REBASEs.  I added a distinct message for the handshake, since if we're going to have a `switch` we might as well use it.

```
$ ./spipe/spipe -t 127.0.0.1:8001 -k cpusupport-config.h 
spipe: Handshake failed
$ echo $?
1
$ ./spipe/spipe -t 127.0.0.1:8002 -k /dev/null 
spipe: Could not connect
$ echo $?
1
```

gperciva:
I debated adding a prompt about the keyfile, i.e.
```
$ ./spipe/spipe -t 127.0.0.1:8001 -k cpusupport-config.h 
spipe: Handshake failed (are you sure that you have the correct keyfile?)
```

but all the wordings I came up with seemed a bit patronising.  I also figured that the target audience of `spiped` would understand what "Handshake failed" meant.

cperciva:
The code looks good; please rebase.

At some point we might want to distinguish between network errors and cryptographic errors, but I think it's worth getting what we have here out first.

(Also possibly worth having spiped report on connection failures too -- although right now such warnings would probably end up at /dev/null in almost all cases.  Maybe some day we should add logging though...)

gperciva:
Rebased, ready for merge.

gperciva:
Sorry, ignore the message about this being ready..  I was looking at the wrong tab.

gperciva:
Rebased, ready for merge.

PR #205, opened 2019-08-15
Update changelog

gperciva:


gperciva:
Not certain if you want to call this "Unreleased version", or "spiped-1.6.1", or "spiped-1.7.0".  In the `tarsnap` client, we use `(unreleased)`, but the past history of this file in spiped suggests giving it a version number immediately.

CHANGELOG
```diff
@@ -1,3 +1,9 @@
+spiped-1.6.1
+* New option -u username:groupname (spiped): change the user and/or group
+  ownership of the process.
+* Use RDRAND as an additional source of entropy on CPUs which support it.
+* Use SHANI instructions on CPUs which supports them.
```
cperciva: s/supports/support/

gperciva:
Fixed

CHANGELOG
```diff
@@ -1,3 +1,9 @@
+spiped-1.6.1
+* New option -u username:groupname (spiped): change the user and/or group
+  ownership of the process.
+* Use RDRAND as an additional source of entropy on CPUs which support it.
+* Use SHANI instructions on CPUs which support them.
+* Warn about an inability to establish a connection in spipe.
```
cperciva: Since we warn about more than merely failures to *connect* to the target, maybe
"Warn about failed connections and exit with non-zero status (spipe)." ?

(I think the change in exit status is important enough to mention, too.)

PR #207, opened 2019-10-01
Test pushbits

gperciva:


tests/pushbits/main.c
```diff
@@ -76,6 +79,90 @@ echo()
...
+	ssize_t r;
+
+	/* Initialize output message buffer. */
+	if ((buf = malloc(msglen)) == NULL) {
+		warnp("malloc");
+		goto err0;
+	}
+
+	/* Create socket pairs for the input and output. */
+	if (socketpair(AF_UNIX, SOCK_STREAM, 0, in)) {
```
gperciva: Draft PR so that I can ask a question; I'm missing something fundamental about sockets.

- I first attempted to use `socket()` to create two sockets, then connect them with `pushbits()`.  That didn't work because it complained about not being connected.  IOW, I tried to do:

    ```
    write(s0, ...) -> pushbits(s0, s1) -> read(s1, ...)
    ```

- I now have it working with `socketpair()`, but this feels like I have too many sockets / file descriptors floating around.  I now have:

    ```
    write(in0, ...) -> socketpair(in0, in1) -> pushbits(in1, out0) -> socketpair(out0, out1) -> read(out1, ...)
    ```

spipe/pushbits.c
```diff
@@ -70,14 +70,13 @@ workthread(void * cookie)
 }
 
 /**
- * pushbits(in, out):
+ * pushbits(in, out, thr):
  * Create a thread which copies data from ${in} to ${out}.
```
cperciva: Need to adjust the comment to say what `thr` is.  ("... and store the thread ID in ${thr}" perhaps).

spipe/pushbits.h
```diff
@@ -1,10 +1,12 @@
 #ifndef _PUSHBITS_H_
 #define _PUSHBITS_H_
 
+#include <signal.h>
```
cperciva: I guess this works, but I'd say that `<sys/types.h>` is a more normal place to find `pthread_t`.

tests/pushbits/main.c
```diff
@@ -76,6 +79,90 @@ echo()
...
+	/* Link the two pairs. */
+	if (pushbits(in[1], out[0], &thread)) {
+		warnp("pushbits");
+		goto err1;
+	}
+
+	/* Wait for thread to start. */
+	wait_ms(100);
+
+	/* Send the message. */
```
cperciva: Maybe worth adding a comment saying that this will not block because the message is small enough to be entirely buffered within the kernel.  If `msglen` was too large we could block forever since nobody is reading the output yet.

tests/pushbits/main.c
cperciva: This is correct.  For "socket" think "half of a connection"; a socketpair is a complete connection.  So with two socketpairs you can be writing into one connection, pushbits takes data out of that connection and pushes it into the other, and then you read data out of the second connection.

gperciva:
New version; still a draft.

You mentioned (I think in the car) that we should test that we can cancel one pushbits() thread without disrupting the other one.  I added a "send a message through two pushbits() threads", but I don't think that's exactly what you meant.  I tried to send a message through the `middle` socketpair after closing the `in`socketpair, but that always complained about the socket not being connectect.  Is that what you had in mind, or should I be sending a message large enough that it wouldn't be buffered in the kernel, or...?

cperciva:
The case I'm concerned about is `cat foo | spipe spiped.echo.server > bar` where `spiped.echo.server` is spiped in front of an echo server.

When `cat` finishes working its way through the file and `spipe` reads the EOF, it will close the *sending* side of the TCP connection to the remote `spiped` process, but the *receiving* side needs to keep on working so that we can read the data on its way back.

Easiest way to test this is with a large `foo` file and an echo server which is rate-limited, to ensure that `spipe` notices the EOF while there is still data in network buffers.

gperciva:
Updated test; ready for evaluation.  The "scrypt echo server test" was added in #209.

PR #215, opened 2019-10-22
tests: add valgrind

gperciva:


gperciva:
I'm not certain if we want to merge this now.  It's currently an "aspirational" PR, in that `make test USE_VALGRIND=1` currently fails.  (Note that this PR does **not** add it to Travis-CI, for precisely that reason.)

I can see the argument for delaying until the fixes are in.  OTOH, it would be personally convenient to merge it now, so I don't need to keep on juggling branches to test fixes with valgrind and then rewrite history to remove the valgrind commit so that the "fix xyz" branches are purely the "fix" and not "fix plus the 5-day-old-add-valgrind-tests-that-now-doesn't-rebase-onto-the-lastest-test-suite".

gperciva:
An added complication is that it currently fails on FreeBSD due to an uninitialized variable in `libc`:

```
==75176== Conditional jump or move depends on uninitialised value(s)
==75176==    at 0x4A3E267: ??? (in /lib/libc.so.7)
==75176==    by 0x7FF00019F: ???
==75176==    by 0x49DA538: ??? (in /lib/libc.so.7)
==75176==    by 0x4: ???
==75176==    by 0x4948D0E: ??? (in /lib/libc.so.7)
==75176==    by 0x7FF0001CF: ???
==75176==    by 0x209355: ??? (in /data/td/src/spiped/tests/dnsthread-resolve/dnsthread-resolve)
...
==75176==  Uninitialised value was created by a stack allocation
==75176==    at 0x20B040: ??? (in /data/td/src/spiped/tests/dnsthread-resolve/dnsthread-resolve)
```

On Linux, I'd install `libc-dev` to get the debugging libraries to see which function(s) contain the error (likely `gethostbyname`).  Is there such a package on FreeBSD?  Or is that already installed but the old FreeBSD port of valgrind fails to find the symbols for some reason?

cperciva:
> On Linux, I'd install `libc-dbg` to get the debugging libraries to see which function(s) contain the error (likely `gethostbyname`). Is there such a package on FreeBSD? Or is that already installed but the old FreeBSD port of valgrind fails to find the symbols for some reason?

The default install has debug symbols in /usr/lib/debug/lib/ but those are optional so it's possible that you didn't enable debug symbols at install time.

My guess is that valgrind isn't looking in the right place.

gperciva:
```
$ ls /usr/lib/debug/
boot
$
```

If there was any "customize your install" page during the freebsd install, I would have deselected almost as much as possible because I'd assume that I could install them easily later.  (Also, at the time I was assuming that I'd use this server as a launchpad for virtual machines, not a devel machine in its own right.)

cperciva:
I'm fine with merging this now; if someone says "hey, the test fails if I turn valgrind on" we can always say "yeah, known issues, we're working on it".  Up to you.

cperciva:
> If there was any "customize your install" page during the freebsd install, I would have deselected almost as much as possible because I'd assume that I could install them easily later. (Also, at the time I was assuming that I'd use this server as a launchpad for virtual machines, not a devel machine in its own right.)

Ok.  They can be downloaded and extracted after the fact, but it's probably easiest just to use a virtual machine (e.g., an EC2 instance).

gperciva:
Let's go for merging it, but I've removed it from the README to reduce the chance of somebody running it accidentally.

gperciva:
Yeah, trying it on EC2 was definitely the right call.

PR #216, opened 2019-10-23
Fix pushbits cleanup

gperciva:


gperciva:
Another "maybe" PR.  This fixes the valgrind tests on `tests/pushbits`.

However, when I combine this with `pthread_join()` in `spipe/main.c`, most of the time it works great.  But maybe 10% of the time when I test `spipe ... ^C` I get a segfault.

I'm not certain if that's an existing problem, or caused by this code, or what.  I'm still investigating, so I'm hesitant to put this forwards as a "yes, I think this is ready PR".  OTOH, I can't see how this PR might cause a segfault that didn't already exist.

gperciva:
Good to merge, I think.  I documented my (failed) attempt to stop the pushbits() thread with `close()` in the commit message.

Valgrind doesn't report any errors on FreeBSD, but on Linux it complains about `./test_pushbits 2`, namely the `echo_stdin_stdout()`.  AFAIK that's the intended way of interrupting such a workthread on POSIX systems.

This particular PR doesn't change `spipe`; that will come later.  This one only allows `tests/pushbits` to pass the valgrind tests.

PR #217, opened 2019-10-24
Valgrind freebsd strlen

gperciva:


gperciva:
This allows `make test USE_VALGRIND=1` to work on FreeBSD 12.0 on amd64.

The first commit is importing an improvement from libcperciva.  The second commit is the "real" one, and provides a good test case if you wanted to discuss it with other people.  (I don't think that FreeBSD is doing anything wrong; the only problem is that the valgrind port is really old, but if nobody is interested in maintaining it then that's how it goes.)

PR #218, opened 2019-10-25
Build cleanup

gperciva:


gperciva:
Trivial updates to the build system; absolutely no change when running `make Makefiles`.

PR #222, opened 2019-10-29
Spipe cleanup and valgrind

gperciva:


spipe/main.c
```diff
@@ -212,6 +214,16 @@ main(int argc, char * argv[])
...
+	/* Wait for threads to finish. */
+	if ((rc = pthread_join(threads[0], NULL)) != 0) {
+		warn0("pthread_join: %s", strerror(rc));
+		goto err2;
+	}
+	if ((rc = pthread_join(threads[1], NULL)) != 0) {
+		warn0("pthread_join: %s", strerror(rc));
+		goto err2;
+	}
+
```
cperciva: These will end up calling `sock_addr_freelist(sas_t);` if they fail.  Is that correct?

gperciva:
This now depends on #224 and #226.

gperciva:
@TimWolla reminded us that this is similar to #44.  In particular, @cperciva had some concerns about what if we're writing to a process which stalled https://github.com/Tarsnap/spiped/pull/44#issuecomment-210781703.

gperciva:
@TimWolla reminded us that this is similar to #44.  In particular, @cperciva had some concerns about what if we're writing to a process which stalled https://github.com/Tarsnap/spiped/pull/44#issuecomment-210781703.

gperciva:
@TimWolla reminded us that this is similar to #44.  In particular, @cperciva had some concerns about what if we're writing to a process which stalled https://github.com/Tarsnap/spiped/pull/44#issuecomment-210781703.

spipe/main.c
gperciva: fixed in #224.

gperciva: Or rather, fix *proposed* in #224.

gperciva: Or rather, fix *proposed* in #224.

gperciva:
Ready for merge.

PR #224, opened 2019-10-30
spipe: cleaner handling of sas_t for error path

gperciva:


gperciva:
This arose from #222.

spipe/main.c
```diff
@@ -194,6 +194,9 @@ main(int argc, char * argv[])
 		goto err2;
 	}
 
+	/* proto_conn is now maintaining sas_t. */
```
cperciva: "sas_t is now owned by proto_conn".

(Just a matter of normal jargon; we usually talk about "owning" a memory allocation rather than "maintaining" it.)

PR #226, opened 2019-11-05
spipe: use an "events & threads" cookie

gperciva:


spipe/main.c
```diff
@@ -84,9 +89,8 @@ main(int argc, char * argv[])
 	struct proto_secret * K;
 	const char * ch;
 	int s[2];
-	int conndone = 0;
 	void * conn_cookie;
-	pthread_t threads[2];
+	struct events_threads ET;
```
cperciva: This should be at the top of the "working variables" section.  The compiler might move it anyway; but it's an old UNIX tradition that structures go first in order to avoid creating stack holes in the event that they have stricter alignment requirements.

PR #227, opened 2019-11-05
Spipe graceful shutdown upon SIGTERM

gperciva:


gperciva:
Cleanly shut down `spipe` upon `SIGTERM`.

Depends on #222.

TimWolla:
See also #44

gperciva:
Hmm, I'd say that #222 is more similar to #44.  This particular PR is about sending a `SIGTERM` to `spipe`, so I think it's reasonable to drop any data that's potentially left in output buffers.  (I welcome other opinions, though!)

TimWolla:
Ah, sorry. I didn't realize that these particular commits are part of 222 as well. Just saw that pthread_join stuff and remembered the old one.

gperciva:
Thanks for the reminder, though!  That old PR sparked my interest in this (months or years ago), but I'd forgotten about it when I rediscovered my old branches.  It's particularly good to remind @cperciva about his earlier concerns when he looks at that PR.  :)

cperciva:
Hmm.  I think my concern was that we could end up with `spipe` lingering for an indefinite time due to a write(2) call in pushbits blocking.  Now that we're using pthread_cancel, I don't think that's an issue -- at least, not for the case of "trying to kill the process".  And if we're not explicitly trying to kill the process, I think lingering until it manages to flush its data is the right answer anyway.

So I think this is fine...

gperciva:
Ready for merge.

PR #228, opened 2019-11-06
Misc minor fixes

gperciva:


gperciva:
I find it mildly off-putting that one of the `/* Command-line parameters */` is actually allocated memory (via `strdup`) rather than merely a `const char *` pointing to the relevant part of `argv`.

I have a vague recollection that I tried it once before and you didn't like it, but I can't find  any old PR that it might have been part of.

spiped/main.c
```diff
@@ -241,10 +241,15 @@ main(int argc, char * argv[])
 
 	/* Figure out where our pid should be written. */
 	if (opt_p == NULL) {
```
cperciva: I'm tempted to write this now as

```
if (asprintf(&pidfilename, opt_p ? "%s" : "%s.pid", opt_p ? opt_p : opt_s)) {
    warnp("asprintf");
    goto err0;
}
```

but maybe you'll think that's too obscure?

gperciva: I had something similar to that in a draft, but I did think it was unnecessarily un-obvious.  But if you prefer it that way, I won't object.  I'll upload a new draft.

gperciva:
New draft.  I used the `(opt_p != NULL) ?` form because I'm not comfortable using `opt_p ?` by itself.

PR #229, opened 2019-11-06
Tests fewer sleep

gperciva:


gperciva:
Avoid most of the hard-coded `sleep(1)` commands in the test suite.

Future idea: it might be nice to make a `msleep` utility so that `wait_for_file()` could check every 100ms instead of every 1000ms.

cperciva:
FWIW, FreeBSD's `sleep(1)` can take fractional values.  POSIX doesn't support that, though, and I don't know if any other operating systems follow our example.

PR #232, opened 2019-11-08
Fix header includes for proto_crypt.c

gperciva:


gperciva:
Really fun bug!  See commit message.

Big props to clang here.  The default NetBSD compiler gave me:

```
In file included from ../proto/proto_crypt.c:11:0:
../libcperciva/util/sysendian.h:8:17: error: redefinition of 'libcperciva_be16enc'
 #define be16enc libcperciva_be16enc
                 ^
../libcperciva/util/sysendian.h:8:17: note: previous definition of 'libcperciva_be16enc' was here
 #define be16enc libcperciva_be16enc
                 ^
../libcperciva/util/sysendian.h:29:1: note: in expansion of macro 'be16enc'
 be16enc(void * pp, uint16_t x)
 ^
```

which made me really confused.  ("How can it be complaining about it being previously defined on that same line?!  Could it be ignoring the `#ifndef`?!?  Why can't I reproduce this behaviour in the preprocessor with stand-alone files?!?!?")

By contrast, when I was approaching my second hour of confusion and installed `clang`, I got:

```
In file included from ../proto/proto_crypt.c:14:
In file included from ../proto/proto_crypt.h:5:
In file included from /usr/include/unistd.h:70:
In file included from /usr/include/sys/types.h:98:
In file included from /usr/include/machine/endian.h:3:
/usr/include/sys/endian.h:203:1: error: redefinition of 'libcperciva_be16enc'
__GEN_ENDIAN_ENC(16, be)
^
/usr/include/sys/endian.h:196:31: note: expanded from macro '__GEN_ENDIAN_ENC'
static __inline void __unused \
                              ^
<scratch space>:101:1: note: expanded from here
be16enc
^
../libcperciva/util/sysendian.h:8:17: note: expanded from macro 'be16enc'
#define be16enc libcperciva_be16enc
                ^
../libcperciva/util/sysendian.h:29:1: note: previous definition is here
be16enc(void * pp, uint16_t x)
^
../libcperciva/util/sysendian.h:8:17: note: expanded from macro 'be16enc'
#define be16enc libcperciva_be16enc
                ^
```

which sent me in exactly the right direction.

cperciva:
Looks to me like this is a NetBSD bug.  Unless we're not defining `_POSIX_C_SOURCE` or `_XOPEN_SOURCE` properly?

https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html

gperciva:
Something weird is happening with `_XOPEN_SOURCE`.  It's totally fine with `_POSIX_C_SOURCE`!

I'm continuing to investigate.

```
$ more foo.c 
static int be16dec() { return 16; }
#include <sys/types.h>
$ cc -D_POSIX_C_SOURCE=200809L -c foo.c 
$ cc -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE=700 -c foo.c 
In file included from /usr/include/i386/endian.h:3:0,
                 from /usr/include/sys/types.h:98,
                 from foo.c:2:
/usr/include/sys/endian.h:220:1: error: conflicting types for 'be16dec'
 __GEN_ENDIAN_DEC(16, be)
 ^
foo.c:1:12: note: previous definition of 'be16dec' was here
 static int be16dec() { return 16; }
            ^
$
```

gperciva:
This PR definitely fixes a "libcperciva STYLE" bug; we're supposed to include `<unistd.h>` for `ssize_t`, and this should happen before we include any `"*.h"` files. The remaining question is whether this has exposed a NetBSD bug, and I'm leaning towards "no".

### Finding of fact: running
```
cc -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE=700 -E /usr/include/sys/types.h
```
on NetBSD 8.1 shows that it exposes `be16dec` (amongst other symbols) not listed on https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html).

### Question of law

- it's true that the standard about `_POSIX_C_SOURCE` states:

  > `3.` Additional symbols not required or explicitly permitted by POSIX.1-2017 to be in that header shall not be made visible, except when enabled by another feature test macro.
  > https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_02_01_01

- however, the `be16dec` symbol is protected by `#if defined(_XOPEN_SOURCE) || defined(_NETBSD_SOURCE)`.

- XSI seems to suggest that additional symbols are fine:

  > `4.` The system may provide non-standard extensions. These are features not required by POSIX.1-2017 and may include, but are not limited to:
  > ...
  > Additional symbols in standard headers
  > https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap02.html#tag_02

So I don't think that NetBSD is misbehaving.

gperciva:
For the record, if I omit the `-D_XOPEN_SOURCE`, then `be16dec` is not visible.

gperciva:
Also for the record, we added `-D_XOPEN_SOURCE` after 1.6.0.

cperciva:
I think this might still be a NetBSD bug, but the fix is good anyway.  I'll mention it the next time I'm talking to some NetBSD people.

PR #233, opened 2019-11-08
spiped: remove unnecessary header

gperciva:


gperciva:
Unfortunately my investigation with `iwyu` didn't complain about #232 (when I tested it on master, where there's definitely an include-headers bug).  I'll investigate again once they have a release which conatins the bugfix for the "infinite `struct blah;`" problem I reported (now fixed) https://github.com/include-what-you-use/include-what-you-use/issues/682.

PR #236, opened 2019-11-18
BUILDING: warn about potentially broken c99

gperciva:

Issue #235, opened 2019-11-17
Can not make in SUSE/openSUSE


eastspidar:
##### But it can be successful in redhat/centos. 

--------------------------------------------------------------------------------------------------------
```bash
c99  -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE=700 -DCPUSUPPORT_CONFIG_FILE=\"cpusupport-config.h\"  -I.. -I../proto -I../libcperciva/alg -I../libcperciva/datastruct -I../lib/dnsthread -I../lib/util -I../libcperciva/util -I../libcperciva/cpusupport -I../libcperciva/events -I../libcperciva/network -I../libcperciva/crypto  -O2 -c main.c -o main.o
c99: invalid option -- '2'
usage: c99 [-c] [-D name[=value]] [...] [-E] [-g] [-I directory ...]
       [-L directory ...] [-o outfile] [-O] [-s] [-U name ...] operand ...
       where operand is one or more of file.c, file.o, file.a
       or -llibrary
Makefile:42: recipe for target 'main.o' failed
make: *** [all] Error 2
```
----------------------------------------------------------------------------------------

gperciva:
This arose earlier: https://github.com/Tarsnap/spiped/issues/181.  At the time we ended up closing the issue, but I'm downloading openSUSE now to investigate further.

gperciva:
Yes, I can confirm that openSUSE violates the POSIX spec for `c99` https://pubs.opengroup.org/onlinepubs/9699919799/utilities/c99.html.  I'm looking into options.

For now, please use:
```
make CFLAGS=-O
```

cperciva:
Reported to openSUSE: https://bugzilla.opensuse.org/show_bug.cgi?id=1156949

gperciva:
@eastspidar: a better workaround is to use
```
make CC=cc
```

(or `clang` if that's your preference), to avoid the broken `c99`.

gperciva:
@cperciva I don't think there's much more that we can do with this, other than warning people with https://github.com/Tarsnap/spiped/pull/236.  It might be worth leaving this as an open issue for the next few weeks so that if anybody else runs into it they can easily find the answer.

[...]

PR #237, opened 2019-11-18
checking CFLAGS_DEFAULT

gperciva:


gperciva:
Different attempt at resolving #235; use instead of #236.

cperciva:
I'm not sure I like how this is set up.  I was thinking that we could have a `posix-cflags-filter.sh` file which is created which is either empty or changes the value of CFLAGS, then that can be sourced after the `export CFLAGS="$${CFLAGS:-${CFLAGS_DEFAULT}}";` line.

gperciva:
How obvious would the definition of `CFLAGS_DEFAULT` be, though?

I started off with something that occasionally changed the value of `CFLAGS_DEFAULT` (i.e. "if necessary"), but then I remembered that "cute" or "smart" build systems deserve all the hate they get.  I mean, I could definitely imagine myself:
1. compiling on some platform and getting a build failure.
2. change the `CFLAGS_DEFAULT` in `Makefile`.
3. try to compile again.
4. spend the next 20 minutes grasping at increasingly unlikely bugs in the OS, before I noticed that the value of `CFLAGS_DEFAULT` which was getting passed to the subdirectory `Makefile`s wasn't the same as its definition in the top-level `Makefile`.

I think that it would be much safer if there were different variable names for "what the user (or developers) suggests for the defaults" vs. "the default values that we're actually using".  I'm not wedded to `CFLAGS_DEFAULT_CHECKED`; we could change thE variables to `CFLAGS_DEFAULT_PRE` and `_post, or call the first one `_SUGGESTED`, or... well, I'm not fussy about the exact names.

cperciva:
I'm not talking about changing `CFLAGS_DEFAULT`.  I'm talking about printing "WARNING: POSIX violation: make's CC doesn't understand -O2" and then stripping `-O2` from `CFLAGS`.

cperciva:
I was thinking something like the filter-O2 branch I just pushed.  But that might not be enough to unbreak opensuse anyway. :-(

PR #242, opened 2019-12-20
Filter -O2

gperciva:


Makefile
```diff
@@ -49,9 +49,10 @@ cflags-filter.sh:
 		export CC="${CC}";					\
 		cd ${LIBCPERCIVA_DIR}/POSIX;				\
 		command -p sh posix-cflags-filter.sh "$$PATH";		\
-	else								\
-		:;							\
 	fi > $@
+	if [ ! -s $@ ]; then						\
+		printf "#define CFLAGS_OK 1\n";				\
```
gperciva: I know that you were quite happy that `cpusupport-config.h` could act as both a header file and a shell script, so I've followed that here.  But I'm wondering if this message should be `# CFLAGS_DEFAULT are ok` instead, i.e. not keeping this file as a C99 header.  I don't feel strongly either way.

gperciva:
Currently a draft; if you think these changes look ok, I'll rebase them into your single commit and re-submit it as a non-draft.

Makefile
cperciva: I cared about `cpusupport-config.h` being both a header file and a shell script because we needed both and it was important that they were in sync -- making it just a single file solved the "keep in sync" problem.

Since we're only using `cflags-filter.sh` as a shell script, there's no need to worry about that.  I'd probably spell the message `# Compiler understands normal flags; nothing to filter out` though.

cperciva:
Looks good aside from the message we use in the nothing-to-filter case.

gperciva:
Rebased, ready for merge, tested on freebsd 12.1, openSUSE 15.1, NetBSD 8.1.

PR #244, opened 2020-01-08
Add pthread_create_blocking

gperciva:


gperciva:
The more I read the standard, the more confused I am.  The question is

- after `pthread_create(&thr, ...)` returns 0, is `thr` a valid thread ID?

> Upon successful completion, pthread_create() shall store the ID of the created thread in the location referenced by thread.

so... it should be valid?

> There is no requirement on the implementation that the ID of the created thread be available before the newly created thread starts executing.

... ok, so not necessarily valid?

(very next sentence):
> The calling thread can obtain the ID of the created thread through the thread argument of the pthread_create() function, and the newly created thread can obtain its ID by a call to pthread_self().

... so I can get the ID, but it's not necessarily "available" ?

Assuming that I'm correct that the standard does not guarantee a valid thread ID immediately after `pthread_create()`, this wrapper should take care of blocking until the workthread has inidicated that it's ready.  I expect some quibbles about naming (`pthread_create_blocking` vs. `pthread_util_create` vs. `pthread_util_create_wait` vs. whatever).

cperciva:
Which PR does this depend on?

gperciva:
At the time, it depended on #321 and #320.  The latter looks like it isn't related to `pthread_create_blocking()`, but I ran into it when running the tests on github actions.

Now, those PRs have been merged, but I noticed that this PR doesn't cleanly exit in some weird cases (it was something like "if `pthread_mutex_lock` fails").  So it's still not ready for final review, sorry.

cperciva:
I'm not convinced this is the sanest API.  I mean, it's a fun use of function pointers, taking a pointer to a function which itself takes a pointer to a function -- but wouldn't it be simpler to simply take two functions `init` and `run`?

gperciva:
I'll try to change the API as you suggest; hopefully that'll either work, or else let me ask a better-posed question.

gperciva:
Ah right: if we supply an `init` and `run` function, then `pushbits.c` can't use `pthread_cleanup_push()`, because that can either be a function or a macro (in which case, it absolutely requires a corresponding `pthread_cleanup_pop` within the same function).

It might be possible to move the `pthread_cleanup_*` functions into `pthread_create_blocking.c`, and take a third function: `cleanup`.

Faced with that choice a few months ago, I went with the "fun" function pointer solution, since it seemed less intruisive to `pushbits.c`.

I'll proceed with the 3-function version, partly to see if it works, and partly so that you can compare the two APIs.

cperciva:
Just to clarify, I was thinking `pthread_create_blocking(... init, run, cookie)` which results in a thread being created which essentially runs

```
{
    init(cookie);
    signal_pthread_create_blocking_to_return();
    run(cookie);
}
```

and yes, if init == NULL then it would just be a case of "wait until the thread starts running before returning".

PR #245, opened 2020-02-21
tests/pushbits: ignore ESRCH errors from pthread_cancel

gperciva:


gperciva:
Strictly speaking, a thread ID should still valid until pthread_join() (ignoring a few other quibbles).  This was fixed in FreeBSD:
https://lists.freebsd.org/pipermail/svn-src-head/2020-February/133736.html

However, many platforms still return ESRCH in that case, so for compatibility we ignore them.

tests/pushbits/main.c
```diff
@@ -77,8 +77,17 @@ echo_stdin_stdout(void)
...
 	sleep(1);
 	if ((rc = pthread_cancel(thread)) != 0) {
-		warn0("pthread_cancel: %s", strerror(rc));
-		goto err0;
+		/*
+		 * These tests ignore an ESRCH error returned by
+		 * pthread_cancel.  According to the POSIX standard, a
+		 * thread ID should still be valid once it finishes if
+		 * pthread_join() has not yet been called.  However,
+		 * many platforms return ESRCH in the situation.
```
cperciva: in *this* situation?

tests/pushbits/main.c
```diff
@@ -77,8 +77,17 @@ echo_stdin_stdout(void)
 	/* Wait a short while, then cancel the thread. */
 	sleep(1);
 	if ((rc = pthread_cancel(thread)) != 0) {
-		warn0("pthread_cancel: %s", strerror(rc));
-		goto err0;
+		/*
+		 * These tests ignore an ESRCH error returned by
+		 * pthread_cancel.  According to the POSIX standard, a
+		 * thread ID should still be valid once it finishes if
```
cperciva: "once it finishes" is a bit imprecise (does the thread ID finish?) so I'd go with "after pthread_exit has been invoked by the thread".

gperciva:
Updated with the two changes to the comment.

PR #247, opened 2020-03-27
Warn if the maximum number of connections is reached

gperciva:


gperciva:
DRAFT; I haven't added it to the man-page yet.

You've written
> I assumed that people would have some idea of how many connections they would be running simultaneously and set the option appropriately.

but you've also written
> To me, the fact that users would rather pay Amazon to configure OSS for them -- or pay RedHat to figure out how to install it -- is something developers should take as a strong sign that their software is flawed.

I think there's an inherent tension between those two views, which `--warn-nconn` would help to address.  :)

gperciva:
... in fact, I'm not certain why this isn't the default behaviour.  I think we should definitely print a warning if `nconn_max` is reached, but not require an additional command-line argument to do so.

That would also eliminate the need to edit the man-page.  :)

gperciva:
Travis-CI is having some issues, but that doesn't matter for the currently level of feedback.

cperciva:
On second thought, let's make this an unconditional warning.

gperciva:
Updated to be an unconditional warning, and added to the `CHANGELOG` as well.  Ready for merge.

spiped/dispatch.c
```diff
@@ -92,6 +92,12 @@ doaccept(struct accept_state * A)
 {
 	int rc = 0;
 
+	/* Warn about reaching nconn_max. */
+	if (A->nconn >= A->nconn_max) {
+		warn0("Maximum number of connections (%i) reached",
```
cperciva, line 97: What type is `A->nconn_max`, and how should it be printed?

gperciva:
Sorry.  Fixed now.

PR #250, opened 2020-04-16
Fix clang-10 nitpicks

gperciva:


gperciva:
The second commit arises due to clang 10's static analyzer, not the compiler itself.

PR #252, opened 2020-05-02
BUILDING: Solaris utilities might not be POSIX-compatible

gperciva:


gperciva:
If you're curious, here's the man-page:
    https://docs.oracle.com/cd/E88353_01/html/E37853/standards-7.html

I didn't include the URL in `BUILDING` because the url looks prone to breakage.  (Also, we haven't included any other such urls there.)

I have a separate branch which replaces `grep -F "${cmd}"` with `grep "${cmd_escaped}", which works on Solaris, but when I was in the middle of writing the comment about it, I realized that you wouldn't want to change our code to accommodate it.

I wouldn't be surprised if we get bug reports about this in the future, since their man-page isn't an obvious place to check, but hopefully this note in `BUILDING` will help.

cperciva:
This looks plausible, but I'm inclined to wait until you finish figuring out Solaris breakage, unless you're sure there won't be any changes to this.

gperciva:
I'm certain that we have three main options for spiped on Solaris:

1) Leave `make test` broken on Solaris.  I'm not a fan of this, because users shouldn't get into the habit of accepting projects with broken tests.
2) Document that we require POSIX utilities (i.e. this PR).  Based on the language of `standards(7)`, I strongly suspect that this is the option which Solaris developers would suggest.
3) change our test code to avoid `grep -F`, as in https://github.com/Tarsnap/spiped/pull/254.  I don't think that you'll like this.

... I suppose there's a few other options, too:
4) the test suite could detect if the system has `fgrep` or `grep -F`, set `$(GREPF)`, and use that variable in the test suite.
5) rewrite the tests to use `127.0.0.1:8001` instead of `[127.0.0.1]:8001`.
6) I can come up with increasingly ugly solutions if you want more brainstorming.  :)

gperciva:
(I'm also certain that this has nothing to do with the autoconf + `command -p` + Solaris issue.)

PR #254, opened 2020-05-05
tests: avoid needing grep -F

gperciva:


gperciva:
I'm uploading this as additional material for considering https://github.com/Tarsnap/spiped/pull/252; I don't think that we want to accept it.

PR #257, opened 2020-06-24
Libcperciva import

gperciva:


gperciva:
This deliberately does not include the "check that we've created all exitfiles" commit.  There were a few of those omitted in the spiped tests; some by accident, but some because it was a quick hack to make the valgrind tests pass.  Cleaning this up doesn't belong in a "libcperciva import" PR.

PR #261, opened 2020-07-20
Fix misc

gperciva:


spipe/main.c
```diff
@@ -71,8 +71,17 @@ callback_graceful_shutdown(void * cookie)
...
-			goto err0;
+			/*
+			 * According to the POSIX standard, a Thread ID should
+			 * still be valid after pthread_exit has been invoked
+			 * by the thread if pthread_join() has not yet been
+			 * called.  However, many platforms return ESRCH in
+			 * this situation.
+			 */
+			if (rc != ESRCH) {
+				warn0("pthread_cancel: %s", strerror(rc));
```
gperciva, line 82: I'm not certain about whether this would be useful in `err4` and `err5` of spipe.  So far, I haven't seen any cases where we get ESRCH at those locations (whereas I *have* seen it here).

I'm happy to add this in those other locations as well if you want.

cperciva: Assuming you mean err4 and err5 of `main`: Worst case we get a warning printed there, since we don't abort on errors in the error path.  So I wouldn't bother changing it there.

PR #263, opened 2020-07-24
Libcperciva import & fix tests

gperciva:


gperciva:
This is mostly importing from libcperciva; only the very final commit f0912e1 changes spiped-specific sh code.

The delay came from the "sanity test that we've written the previous result", which revealed that valgrind has problems with daemonized programs.  That required a bunch of updates to the test suite to handle such programs, which are all in this PR.

Issue #248, opened 2020-04-08
Optional logging to syslog


cperciva:
We should teach `warnp.c` to (optionally) send output to syslog instead of stderr.  (This belongs in libcperciva since syslog is a POSIX interface.)

If the appropriate command line option is used, the change should occur immediately after daemonizing.

gperciva:
Added in #267.

PR #267, opened 2020-08-03
spiped: add --syslog to send warnp messages to syslog

gperciva:


gperciva:
This was prompted by #248.  However, you said that "the change should occur immediately after daemonizing", which I'm not certain about.  I would have thought that the change should happen immediately above `events_spin()`, i.e. after all the initialization is finished.  For example, would we ever want "Failed to start graceful_shutdown timer" to appear in syslog as opposed to stderr?

This PR switches to syslog just above `events_spin()`.  If you're certain that you want it immediately after daemonizing (in both places), I can change it accordingly.

gperciva:
Another question: I went with `-l` (that's a lower-case L) because all the other spiped options are lower-case letters.  But I'm now thinking that it would be clearer to call it `--syslog`.

(This is particularly because I glanced at the list of PRs, saw a "negative one", and got really confused because I didn't remember doing that.)

cperciva:
Once a program has daemonized, it is very likely that stderr either goes nowhere (someone starts a daemon and then exits their login shell) or goes somewhere actively harmful (someone starts a daemon and then opens an editor; their screen gets partially overwritten by the warning message).

cperciva:
And yes, I agree with `--syslog` for this purpose.

cperciva:
I assume you know what needs to be revised here and you're not waiting for me?

gperciva:
It was mainly waiting on 0b9b5fa.  I also had it as waiting for #272, but in retrospect it doesn't actually need to delay until then.

Ready for merge.

cperciva:
Commit message needs s/-l/--syslog/.

gperciva:
Oops, fixed.  Also speaking of "oops", I just noticed that this branch is called `log-stdout` instead of `log-syslog`.

I'm happy to upload a new branch and create a new PR, or we could just accept that the merge commit message will be a bit misleading.  (We haven't worried about merge commits in the past.)

PR #268, opened 2020-08-10
Don't close input socket in pushbits

gperciva:


cperciva:
I'm not sure why you want to not close stdin.  spipe acts like a filter process, so when its stdout is closed on the other end, it should close its stdin.  e.g. if we have `cat foo | bar | baz | spipe blah blah blah | fooble | more` and hit `q` in the `more` process, the pipe-closing should cascade all the way back to the original `cat`.

Of course if the process is exiting then the descriptor will be closed automatically by the kernel.  But that's a side issue.

gperciva:
This arose while I was investigating #270, which also prompted things like `flockfile()` in warnp and clarifying that proto_conn takes over s[1] in 6da4414.

There's two reasons that I want to remove `close()` from `pushbits()`: a vague "API design clarity / style" feeling, and ease of testing.  Neither reason is airtight.

I'm trying to avoid weird side effects from `pushbits()`, and closing `stdin` seems like a weird side effect.  With one exception [1], none of our other binaries close the standard streams.

Sure, we could document this side effect as we did with socket shutdown() in be96eed.  But it would be helpful for `tests/pushbits/test_pushbits` if we could check all its cases to ensure that there's no file descriptors left open (or unexpectedly closed) with https://github.com/Tarsnap/libcperciva/pull/280.  Whenever anything goes wrong in `spipe`, my first instinct is to assume that `pushbits()` is the problem, so it would be nice if `tests/pushbits/test_pushbits` could check as many aspects of that function as possible.

When `pushbits()` was first written, there was no clean-up attempt; that was added between 1.5.0 and 1.6.0.  Of course spipe should continue to function as a filter, and as you noted, the kernel will close stdin when the program ends.  It's possible that removing the explicit `close()` will uncover some case where thta doesn't happen, but then we can fix whatever underlying bug that exposed.

[1] `kivaloo/tests/s3/main.c`, which replaced `fcloseall()` with three separate `fclose()` commands.  I'm not convinced that this is the best solution for that binary, but that's a separate issue.

gperciva:
For clarity: so far I have no evidence that this would help with #270.  I still think it's a worthwhile change, but if you're not convinced then I'll document the `close()` in the API comment.

gperciva:
Right, I remember now.  The `close()` in `pushbits()` is problematic because it introduces a dependency between the two workthreads.  Namely, if the threads clean up in the wrong order, we get:

```
$ ./spipe/spipe -t /bad -k /dev/null
...
spipe: pushbits workthread_cleanup close P->in 3...
spipe: pushbits workthread_cleanup ... done close P->in 3
...
spipe: pushbits workthread_cleanup shutdown P->out 3...
spipe: Error shutting down socket: Bad file descriptor
```

Admittedly this is after I've added a few `sleep(1)` commands, and triggered `callback_graceful_shutdown(ET)` to force a cleanup as soon as we detect a connection error.  Note, for example, the comment in `callback_graceful_shutdown()`:

```
        /*
         * We need to cancel & join these threads in order, because otherwise
         * we risk thread[1] closing a socket before thread[0] tries to
         * shutdown() that socket.
         */
```
added in 4311604265f2f7b0ad689741b5894e1eca17cb90.

Granted, we could deal with the dependency in other ways -- I'm sure we could play games with mutex or condition variables or something.  But it seems much simpler to remove the `close()`.

Full log:

```
$ ./spipe/spipe -t /bad -k /dev/null
spipe: main socketpair; now have 3 4
spipe: main created pushbits ET.threads[1] 0x800254a00: 3 1
spipe: workthread read P->in 3...
spipe: main created pushbits ET.threads[0] 0x800255400: 0 3
spipe: workthread read P->in 0...
spipe: main did extra sleep to ensure that pushbits() have started
spipe: proto_conn_drop close C->s 4...
spipe: proto_conn_drop ... done C->s 4
spipe: Could not connect
spipe: workthread ... done read P->in 3...
spipe: callback_graceful_shutdown join 0...
spipe: pushbits workthread_cleanup close P->in 3...
spipe: pushbits workthread_cleanup ... done close P->in 3
spipe: pushbits workthread_cleanup shutdown P->out 1...
spipe: pushbits workthread_cleanup ... done shutdown P->out 1
spipe: pushbits workthread_cleanup close P->in 0...
spipe: pushbits workthread_cleanup ... done close P->in 0
spipe: pushbits workthread_cleanup shutdown P->out 3...
spipe: Error shutting down socket: Bad file descriptor
```

spipe/main.c
```diff
@@ -250,7 +250,7 @@ main(int argc, char * argv[])
 	if ((conn_cookie = proto_conn_create(s[1], sas_t, 0, opt_f, opt_g,
 	    opt_j, K, opt_o, callback_conndied, &ET)) == NULL) {
 		warnp("Could not set up connection");
-		goto err2;
+		goto err3;
```
cperciva, line 253: If we fail at this point, isn't `s[1]` leaked?  But in the error path we're only closing `s[0]`...

gperciva: Fixed in a new commit, because technically it's not part of "don't close... in pushbits".

PR #269, opened 2020-08-20
Libcperciva import

gperciva:


gperciva:
This does not include the recent "track fds" changes to the test suite.

PR #272, opened 2020-08-27
Track fds

gperciva:


gperciva:
Ready for merge.  The "check fds as well" is an old commit from libcperciva, which I held back because it needed other commits before it would pass.

Issue #270, opened 2020-08-20
spipe doesn't gracefully handle an initial connection failure


gperciva:
`spipe` appears to hang when it's pointed at an unreachable socket:
```
$ ./spipe/spipe -t /bad-sock-name -k /dev/null
spipe: Could not connect
```

(the same applies for an IP address that isn't in use, such as `192.168.2.123:8001` on my local network)

It's easy to cancel it with `^C`, but it would be nicer if it could quit by itself (with exit code 1, rather than 130 for the `SIGINT` I sent it manualy).

cperciva:
Oops.  It looks like we catch the error as soon as we try to write to the socket we failed to open though:

```
$ spipe -t /nosuchsock -k /COPYRIGHT 
spipe: Could not connect
asdad
spipe: Error writing: Bad file descriptor
```

Clearly something's broken in the error-handling path...

gperciva:
Is the "success instead of ESRC" [1] included in 12.1-RELEASE-p8, or is that only going into 13.0?

[1] https://svnweb.freebsd.org/base?view=revision&revision=357894

gperciva:
Never mind the `ESRCH` question.

gperciva:
Fixed in #274.

PR #274, opened 2020-09-02
Fix spipe apparent "hang" with an unreachable socket

gperciva:


gperciva:
This fixes #270.  The fix works on its own (as it is here), but it produces valgrind leak warnings.  If this is rebased on top of #272, there are no valgrind errors.

gperciva:
I changed my mind because the red "X" didn't look nice.  This is rebased on #272, showing that it doesn't have any leaks once the other fixes are in.

gperciva:
This doesn't actually depend on #272 (track fds), but I wouldn't mind delaying it until then, just for the extra sanity check of passing the extra tests.

PR #285, opened 2020-11-19
cpusupport: add CPUID_COUNT

gperciva:


gperciva:
Big whoops on my part.

This should have been part of
    2018-08-31 libcperciva: import SHANI and SSE3 code
    9c3840e08367857bdf020fe76319abd89dd5a1e9

As a result of missing this file, no SHANI instructions were ever used, as we needed `cpuid_count()` in order to detect whether the CPU supported them.  Unfortunately this was part of the spiped-1.6.1 release, for whom SHANI support was 1 of the 4 news items.

I missed this because very few CPUs support SHANI.  AFAIK, the first "easily accessible" platform which supported those instructions was EC2 t3a (which have AMD EPYC), but there's no mention of SHANI in the announcement so I didn't realize that I could have tested it there (starting 2019-04-24).

Our other software looks ok.

gperciva:
Since you're going to ask: I tested this in a "standalone encryption" program (coming in a separate PR), which just loops `proto_crypt_enc()`.  On t3a, that went from 81.1 MB/s with software SHA256, to 154.7 MB/s with hardware SHANI.

PR #287, opened 2020-11-23
Time stand-alone encryption

gperciva:


gperciva:
This is a preliminary step before adding performance tests for the real programs.  Having a single binary eliminates various uncertainties in timing, and makes it easier to use analysis tools like valgrind.

This binary provides a "theoretical maximum spiped speed" by allowing you to time various parts of the encryption:
- SHA256 alone.  (We might want to remove this; I'm no longer convinced that it's worth including it, because libcperciva includes this literally, and spiped calls HMAC_SHA256.)
- HMAC_SHA256 alone.
- AES-CTR alone.  (I think we should keep this one, even though libcperciva has it as well.)
- AES-CTR followed by HMAC_SHA256.
- Call `proto_crypt_enc()` directly.

As an example of comparing between systems (my physical server, and EC2 t3 and t3a):
- we can see that the biggest bottleneck on most amd64 systems is the (software) SHA256 calculations.
- t3a has hardware SHA256 so that's fast, and an unexpectedly slow AES-CTR; so the latter is the bottleneck.
- in both cases, we can see that "aesctr + hmac_sha256" is approximately equal to those calculations done individually.  (Determined by holding a thumb and forefinger against the aesctr point, then moving my thumb up to the hmac_sha256 point and comparing that with the aesctr+hmac_256 point.)    That's why I thought it was useful to visualize the time, rather than speed.

standalone-enc

As an example of how it could be useful, here's an example of some theoretical improvements to the code (all timed on my server).

standalone-improvements

gperciva:
Caveat: although I'm quite happy with the overall idea here, there's two uncertainties:

- I'm not fond of the name `standalone-enc`, but I'm drawing a blank for anything better.
- as previously mentioned, I'm now leaning towards removing the "SHA256 only" test.

perftests/standalone-enc/main.c
```diff
@@ -0,0 +1,450 @@
...
+#include "proto_crypt.h"
+#include "sha256.h"
+#include "sysendian.h"
+#include "warnp.h"
+
+/* Smaller buffers are padded, so no point testing smaller values. */
+static const size_t perfsizes[] = {1024};
+static const size_t num_perf = sizeof(perfsizes) / sizeof(perfsizes[0]);
+static const size_t nbytes_perftest = 1000000000;	/* 1000 MB */
+static const size_t nbytes_warmup = 100000000;		/* 100 MB */
```
cperciva, line 20: Weirdly inconsistent indentation of the comments here.

perftests/standalone-enc/main.c
```diff
@@ -0,0 +1,450 @@
...
+	/* Inform the user. */
+	printf("%s", str);
+	if (use_hardware_shani)
+		printf(" using hardware SHANI");
+	else
+		printf(" using software SHA");
+	if (use_hardware_aesni)
+		printf(" and hardware AESNI");
+	else
+		printf(" and software AESNI");
```
cperciva, line 47: s/AESNI/AES/

cperciva:
I'm ok with removing SHA256 since, as you say, spiped doesn't actually use it directly.

perftests/standalone-enc/main.c
gperciva: That's a github display thing.  :(

It's something weird like their "display tab" code counts the "+ " as part of the line, or they just have an off-by-one error somewhere.  The actual c file is fine in this respect.

gperciva:
Oh, another uncertainty: this PR exposes `mkkeypair()` in `proto_crypt.h` if we've defined `STANDALONE_ENC_TESTING`.

I imagine that this is ok, but I wanted to make sure you that you noticed this, because it's not usual to change "real" code in a PR that adds a test.

cperciva:
Yeah, I'm ok with `mkkeypair` gaining external linkage for testing.

cperciva:
I think this is fine but the tests aren't happy -- can you figure out what's up with that?

gperciva:
Tests are broken because I cancelled them.  I noticed a bunch of mistakes (the most recent was the "move column extraction into the Makefile"), so I figured that was a cheap way to indicate that the PR wasn't ready.

I've added a bunch of REBASE commits, which I'm happy to condense and prepare for the final merge if you want.

cperciva:
Yep, all those changes look good.  Please rebase.

gperciva:
Ready for merge.

PR #289, opened 2020-12-02
proto_crypt: only initialize HMAC_SHA256 once

gperciva:


gperciva:
There's some weirdness in HMAC-SHA256 times on x86, but I'm not certain it's worth digging in much more since it's using software rather than hardware instrinstics.  I've double-checked that `proto_crypt.o` is the only object file that changes, so I'm guessing that the changes to HMAC-SHA256 arise due to the exact locations in memory and/or what the cache is doing?

First, let's see that it has the expected effect on ARM: a nice -5% to the time spent doing `proto_crypt_enc()`, and no other changes.

NB: these graphs show the *time*, not speed, so lower is better.

### clang ARM

clang-arm

### gcc ARM
gcc-arm

### clang x86 (t3 instance)

Slight (unexpected) increase in HMAC-SHA256 times.

clang-x86

### clang x86 (personal bare-metal i3-7100U)

Unexpected -4.5% drop in HMAC-SHA256, unexpected drop of -6.0% in aesctr, yet an overall increase in aesctr+hmac_sha256.  (Percentages are relative to the previous measurement, not relative to the baseline of `master`)

clang-mine


### gcc x86 (t3 instance)

Wild, unexpected, and unexplained drop of -9.4% in HMAC-SHA256 times.

gcc-x86

### gcc x86 ( bare-metal i3-7100U)

Wild, unepxected, and unexplained increase of +2.1% and then +4.7% in HMAC-SHA256 times.

gcc-mine

### more comments

The two x86 systems both used gcc 9.3.0, but t3 was FreeBSD 12.2 while the bare-metal system was 12.1.  (I don't imagine that makes much difference.)

I did try splitting up this PR into multiple commits, as granular as possible (i.e. one commit to add the context to the struct, another to initialize the context, etc.).  That left me with more questions than I had before -- apparently, initializing the `HMAC_SHA256_CTX ctx` (even when I wasn't using it) gave a statistically significant speed boost in clang, but penalty in gcc.

proto/proto_crypt.c
```diff
@@ -330,6 +334,9 @@ proto_crypt_enc(uint8_t * ibuf, size_t len, uint8_t obuf[PCRYPT_ESZ],
 	/* Sanity-check the length. */
 	assert(len <= PCRYPT_MAXDSZ);
 
+	/* Copy the original (initialized) context. */
+	memcpy(&ctx, &k->ctx_init, sizeof(HMAC_SHA256_CTX));
```
cperciva, line 334: Can you move this down to just before the "Append an HMAC" section?  Otherwise we might load data into the cache and then evict it before it's used.

gperciva: Rebased as requsted.

PR #290, opened 2020-12-05
Libcperciva import

gperciva:


gperciva:
On second thought, let's wait until we have aes-ctr blocks.

PR #293, opened 2020-12-22
libcperciva import valgrind

gperciva:


gperciva:
The final commit (1e97153) is not part of importing, but is necessary for reliable tests.

This PR does not allow us to run the valgrind tests on FreeBSD; that's coming in a second PR once this one is merged.

PR #295, opened 2020-12-31
Coverity nitpicks

gperciva:


gperciva:
Two nitpicks and one real flaw; all in test code.

cperciva:
Wasn't coverity complaining about pthread_join ?

gperciva:
Oh right, oops. I'm used to github, which displays comments below the line, instead of above it.

gperciva:
Updated.  I kept the "check pthread_cancel" part, even though coverity-scan wasn't actually complaining about it, because I imagine that in a few months it would do anyway.

PR #299, opened 2021-01-11
Libcperciva import

gperciva:


gperciva:
In case you're curious about the effect on benchmarks,
compilers

Blue is before, green is after, using `perftest-compilers.py`.  I hid the legend box because it was too big and didn't really add any useful info.

NB: this is showing **time**, not speed, so lower is better.

cperciva:
Is that graph from this PR?  On my laptop I see 74.3 MB/s on Nov 23, 91.6 MB/s before this PR and 95.3 MB/s after this PR.

gperciva:
Yes, the graph is from immediately before this PR, to after.  On the system that generated that timing data, I see 95.9 MB/s before, and 101 Mb/s after.

* oh wait, that's with clang-11.  With the default `cc` in FreeBSD 12.1 -- which is clang-8 -- I see 98.7 after this PR.  FreeBSD 12.2 moved to clang-10 by default.  With clang10, I see 100.7 Mb/s after this PR.

I'll be creating a branch which contains spiped-1.6.1 + this perftest, but without any optimizations, so that it's easy to compare to the previous official release.

PR #302, opened 2021-02-10
Libcperciva import & adjust for sha256_sse2

gperciva:


gperciva:
The cflags in a415f4b will be removed in the following "libcperciva import" PR (which will be the "extract CPUSUPPORT CFLAGS from .c files").  But I wanted this particular PR to be focused on `sha256_sse2`, so that meant using the old infrastructure.

PR #303, opened 2021-02-10
Libcperciva import & reinstate manual cflags-

gperciva:


gperciva:
2a94568 was necessary to avoid changes to the generated `Makefile` in `perftests/standalone-enc`.

PR #307, opened 2021-03-09
Allow compiling perftests/standalone-enc with bmake

gperciva:


gperciva:
This is a preliminary step which arose from https://github.com/Tarsnap/libcperciva/issues/383#issuecomment-793139151.

Once it's merged, I have the "share object files unless we have directory-specific overrides" branch ready.

gperciva:
Just to confirm: this doesn't change the generated makefiles on FreeBSD or linux.

PR #308, opened 2021-03-09
Build with shared object files

gperciva:


gperciva:
I've got branches doing this for all the relevant Tarsnap projects.  If you want to apply this to any private projects which use the same build system, I've attached a quick python script.

```
find . -name "Makefile.BSD" | xargs ./update-shared.py
```

PS: rename it to remove the `.txt`, which I had to add because github is like that.
[update-shared.py.txt](https://github.com/Tarsnap/spiped/files/6112072/update-shared.py.txt)

gperciva:
8058ff6 was necessary because FreeBSD was expanding SRCS, but Ubuntu wasn't.  Fortunately, it was documented that `-V` depends on a bunch of stuff while `-v` always expands, so the solution was easy.

cperciva:
This does indeed solve the problem I described.  I'm not sure I like it though.  Can we generate a single `lib.a` archive file and have programs pull the bits they need out of that?  I think it would simplify the `Makefile.BSD` files too since they wouldn't need to enumerate all the `.c` files they use.

gperciva:
I've got a branch which appears to work with a library, but I can't find reference(s) in any spec to say that it should always work.  The issue is our ability to override symbols in a library -- or rather, if we have a symbol in an object file, are guaranteed that the linker will not search for that symbol in a library?

For example, 
```
c99 main.o proto_crypt.o -lspiped -L../../lib
```

`libspiped.a` contains the usual (compiled) proto_crypt.c, but without the `-DSTANDALONE_ENC` define, so it doesn't expose `mkkeypair`.  We need that, so I've compiled `proto_crypt.o` independently with that define.  But that means that we'll have (for example) an external symbol of `proto_crypt_secret` in `proto_crypt.o`, and another external symbol of `proto_crypt_secret` in `libspiped.a`.

The man page for `ld`, and all sorts of blog posts and forum answers about the linker say that the above command-line should work.  For example, from `man ld` (on freebsd):
> When ld.lld encounters an undefined symbol that can be resolved by an object file contained in a previously processed archive file, it immediately extracts and links it into the output object.

However, `ld` is (surprisingly!) not part of POSIX, nor can I find anything about the order of resolving symbols during linking in c99.

It's unambiguous that this works with the clang and gcc toolchains.  And there's a whole raft of adjacent issues; e.g., solaris and gcc have "weak symbols" vs. "strong symbols".  But I was hoping to find a definitive reference about it.  :(

Maybe the "the linker searches object files and libraries in the order specified on the command line" is so simple that it's a trivial assumption, which doesn't need to be in any spec?

cperciva:
POSIX talks about linking to multiple `.a` files and how to ensure that the (c99) linker gets the version of a symbol which you want (put it earlier in the command line) so I think it's safe to assume the same applies to having severl `.c` files and a `.a` file.

Rather than using `-L` and `-l` I would just specify the path explicitly, i.e. put `../lib/libspiped.a` into the `c99` command line.

PR #309, opened 2021-03-17
Build with a static library

gperciva:


gperciva:
Nope, just discovered a big problem: it doesn't automatically track the dependencies of the static library.  Closing this for now.

cperciva:
One thing I noticed looking at the changes: I'm guessing we can drop some of the include directories from the build for final binaries -- some might have only been needed for one library file to be able to use another library file.

gperciva:
Oh, good point!

gperciva:
BTW, I just noticed that we have `.PHONY:` in the top-level `.POSIX:` `Makefile` (which I added in b829061916f3c6bc1edc81d18f24887b7a25f411).

That's not part of POSIX yet so it's technically a bug, but it was accepted on 2020-03-18:
https://www.austingroupbugs.net/view.php?id=523
so I suggest that we leave it there.

gperciva:
Ok, I believe that `perftests/standalone-enc/` is working properly.  I can touch `libcperciva/crypto/crypto_aes.c` (or `libspiped/libspiped.a`), and `make` in that directory does the right thing.

I can apply that fix to the other directories, as well as the general clean-up I mentioned earlier.

libspiped/Makefile.BSD
```diff
@@ -0,0 +1,102 @@
...
+
+# Don't install it.
+NOINST	=	1
+
+# Useful relative directories
+LIBCPERCIVA_DIR	=	../libcperciva
+LIB_DIR		=	../lib
+PROTO_DIR	=	../proto
+
+# spipe protocol
```
cperciva, line 12: s/spipe/spiped/

It's the `spiped` protocol even if it's being used by `spipe`.

cperciva: Also, we should probably move `proto` to `lib/proto`.  But let's do that after we finish reorganizing the build code.

release-tools/Makefile.prog
```diff
@@ -26,6 +26,6 @@ install:${PROG}
 clean:
 	rm -f ${PROG} ${SRCS:.c=.o}
 
-${PROG}:${SRCS:.c=.o}
-	${CC} -o ${PROG} ${SRCS:.c=.o} ${LDFLAGS} ${LDADD_EXTRA} ${LDADD_REQ} ${LDADD_POSIX}
+${PROG}:${SRCS:.c=.o} ${PROJECT_LIBS}
```
cperciva, line 29: Don't you need to add code to `Makefile.prog` telling it how to build `${PROJECT_LIBS}`?

gperciva: That was my first thought as well, but then I realized that I could rely on the "loop-back" building, just like `cpusupport-config.h`.

Namely, when you type `make` in a subdirectory:
- it jumps back to the top level and does `make buildsubdir` (with some variables set)
- `buildsubdir` checks if it needs build `cpusupport-config.h` etc.
- `buildsubdir` also unconditionally does `cd libspiped && make all`, which ensures that `libspiped.a` is up to date.
- then it returns to the subdirectory and does `make $PROG`.

(as a side note, I just realized that $PROJECT_LIBS won't work with `make -f Makefile.BSD` as it is currently, but it only needs a small fix.)

((even after that fix, `make -f Makefile.BSD` won't track libspiped.a as a dependency, but hopefully that's ok.))

cperciva: Ah, got it.  So effectively `HAVE_BUILD_FLAGS` is `HAVE_BUILD_FLAGS_AND_LIBFILE_IS_UP_TO_DATE`.  (No, I'm not suggesting that the variable name gets changed.)

cperciva:
Aside from the s/spipe/spiped/ comment I think this is good.  Go ahead and clean up the history and I'll review it again.

gperciva:
Updated, splitting the changes up more.  Now there's three part, with a "rebuild makefiles" at the end of each part.  This provides a nice sanity check, as well as being a template for importing these changes to other projects.

1. build libspiped.a.
2. link binaries to libspiped.a.
3. stop building object files in other subdirectories.

(Wow, the github actions finished much faster!  Looking forward to adding this to kivaloo!)

Makefile
```diff
@@ -12,7 +13,9 @@ TEST_CMD=	tests/test_spiped.sh
 
 ### Shared code between Tarsnap projects.
 
-all:	cflags-filter.sh cpusupport-config.h posix-flags.sh
+LIBFILE=${LIB}/${LIB}.a
```
gperciva, line 16: This is the last quibble that I'm aware of: the top-level `Makefile` only works if there's a single library.

Supporting multiple LIBFILES woudn't be hard -- just move this line above the "Shared code" comment, and make it
```
LIBFILES=libspiped/libspiped.a
```

That does mean there's a bit of duplication with the earlier `LIB`, but given the extra flexibility in terms of allowing multiple libraries, it might be worth it.

(I did consider splitting `libspiped.a` into `libcperciva.a` and `libspiped.a`, where the latter only containing graceful_shutdown, dnsthread, and proto.  In the end I figured against it, but only a 30-70 split.  In our projects, I don't think it really matters either way, but an ideal build system should be able to scale up to huge projects, in which case splitting libraries would be more useful.)

release-tools/Makefile.prog
```diff
@@ -26,6 +26,6 @@ install:${PROG}
 clean:
 	rm -f ${PROG} ${SRCS:.c=.o}
 
-${PROG}:${SRCS:.c=.o}
+${PROG}:${SRCS:.c=.o} ${PROJECT_LIBS}
 	${CC} -o ${PROG} ${SRCS:.c=.o} ${LDFLAGS} ${LDADD_EXTRA} ${LDADD_REQ} ${LDADD_POSIX}
```
cperciva, line 30: I think it would be better to put `${PROJECT_LIBS}` here (before `${LDFLAGS}`) rather than including it in `LDADD_REQ`.  It shouldn't matter but theoretically it might if someone has an `LDADD_EXTRA` which conflicts with a symbol we use.

gperciva: I had that a few hours ago, but then I discovered that it broke `make -f Makefile.BSD`.

Makefile
cperciva: I don't think there's any point splitting the library.  If anything I would consider going in the opposite direction and making this a hard-coded `liball.a` -- partly because that would remove boilerplate of every `Makefile.BSD` specifying that yes, it uses library code, and partly to reduce the risk of someone misunderstanding the build process and thinking that "libspiped" is something which should be installed.

release-tools/Makefile.prog
cperciva: Oh, hmm.  That could be solved by adding it to `LDADD` in `Makefile.inc` I think?

Makefile
gperciva: If we can drop `make -f Makefile.BSD` support, then I'm all for `liball.a`.

... hmm, I suppose that you could do:
```
make -f Makefile.BSD LDADD_REQ="../../libspiped/libspiped.a -lcrypto"
```
(in the example of perftests/standalone-enc)

cperciva: Sounds like another reason to put it into `LDADD` in `Makefile.inc`?

release-tools/Makefile.prog
gperciva: Huh, I forgot about that file.  If nothing else, that would be a better place to stuff `libspiped.a`.

I'll revise this PR with `liball.a` and `Makefile.inc`.

gperciva:
This is ready for another review.

tests/valgrind/Makefile
```diff
@@ -18,8 +18,8 @@ all:
 clean:
 	rm -f ${PROG} ${SRCS:.c=.o}
 
-${PROG}:${SRCS:.c=.o}
-	${CC} -o ${PROG} ${SRCS:.c=.o} ${LDFLAGS} ${LDADD_EXTRA} ${LDADD_REQ} ${LDADD_POSIX}
+${PROG}:${SRCS:.c=.o} ${LIBALL}
+	${CC} -o ${PROG} ${SRCS:.c=.o} ${LIBALL} ${LDFLAGS} ${LDADD_EXTRA} ${LDADD_REQ} ${LDADD_POSIX}
```
gperciva, line 22: These changes are not desired, but end up being harmless, because we haven't defined ${LIBALL} in this `Makefile`. 
 The reason they appeared: `${LIBALL}` is now part of `Makefile.prog`.  Due to the `NOLIBALL = 1` in `Makefile.BSD`, we haven't defined that variable, but we're still copying `Makefile.prog` verbatim.

Technically, things would get confused if somebody has a `LIBALL` environment variable, but I think that's sufficiently unlikely that it's not worth complicating the build to strip it from this one test directory.

liball/Makefile
```diff
@@ -0,0 +1,124 @@
+.POSIX:
+# AUTOGENERATED FILE, DO NOT EDIT
+LIB=liball.a
+SRCS=sha256.c sha256_arm.c sha256_shani.c sha256_sse2.c cpusupport_arm_sha256.c cpusupport_x86_aesni.c cpusupport_x86_rdrand.c cpusupport_x86_shani.c cpusupport_x86_sse2.c cpusupport_x86_ssse3.c crypto_aes.c crypto_aes_aesni.c crypto_aesctr.c crypto_aesctr_aesni.c crypto_dh.c crypto_dh_group14.c crypto_entropy.c crypto_entropy_rdrand.c crypto_verify_bytes.c elasticarray.c ptrheap.c timerqueue.c events.c events_immediate.c events_network.c events_network_selectstats.c events_timer.c network_accept.c network_connect.c network_read.c network_write.c asprintf.c daemonize.c entropy.c getopt.c insecure_memzero.c monoclock.c noeintr.c perftest.c setgroups_none.c setuidgid.c sock.c sock_util.c warnp.c dnsthread.c graceful_shutdown.c proto_conn.c proto_crypt.c proto_handshake.c proto_pipe.c
+IDIRS=-I../libcperciva/alg -I../libcperciva/cpusupport -I../libcperciva/crypto -I../libcperciva/datastruct -I../libcperciva/events -I../libcperciva/network -I../libcperciva/util -I../lib/dnsthread -I../lib/util -I../proto
+LIBALL=../liball/liball.a
```
gperciva, line 6: This line is not intended, and I think it's worth getting rid of it.

tests/valgrind/Makefile
gperciva: I think I changed my mind -- I'll see about removing these instances of `${LIBALL}`.

gperciva:
This is ready for another review, again.

gperciva:
Mistake in the second-last commit messages (mentions libspiped instead of liball).

perftests/standalone-enc/Makefile.BSD
```diff
@@ -14,51 +14,15 @@ PROTO_DIR	=	../../proto
 # Main test code
 SRCS	=	main.c
 
-# spipe protocol
+# spiped protocol
```
cperciva, line 17: This probably deserves an extra comment pointing out that this isn't a mistake and we're deliberately recompiling because we want different compiler options from what was used in liball.a.

cperciva:
Looks good aside from the commit message you noticed and needing a bit of explanation added to standalone-enc/Makefile.BSD.

perftests/standalone-enc/Makefile.BSD
gperciva: Changed it to:
```
# spiped protocol, with an extra -DSTANDALONE_ENC_TESTING to enable mkkeypair()
```

gperciva:
Updated, hopefully ready for final merge.

PR #310, opened 2021-03-22
Print hardware

gperciva:


gperciva:
Adding the `#if defined(CPUSUPPORT_CONFIG_FILE)` might be overkill, but it's only test code, and it might stop somebody going "huh?!  how is this so fast if it's not using hardware acceleration?  And for that matter, why isn't it using hardware acceleration?!".

gperciva:
This is more relevant once we have a `liball.a` -- in that case, it's quite plausible that `liball.a` could be built with `CPUSUPPORT_CONFIG_FILE`, but if you then use `make -f Makefile.BSD` in the subdirectory, the binary wouldn't know anything about the cpusupport.

PR #316, opened 2021-07-03
Benchmark proto_pipe

gperciva:


gperciva:
To check spiped performance, this adds `proto_pipe()` to `standalone-enc`:
```
$ ./test_standalone_enc 4
Testing spiped speed limits using hardware SSE2 and hardware AESNI.
Testing proto_crypt_enc()
97656 blocks of size 1024	0.709677 s	140.908813 MB/s
$ ./test_standalone_enc 5
Testing spiped speed limits using hardware SSE2 and hardware AESNI.
Testing proto_pipe()
97656 blocks of size 1024	1.255581 s	79.644200 MB/s
```

I was expecting the event loop / network writes to slow things down, but not (almost) cutting the throughput in half.  I'm quite willing to believe that I'm doing something silly in ecb2d36.

If you like this general idea, then I'll split off two previous PRs (improving `perftest_buffers()`, and maybe resurrecting `pthread_create_blocking` from #244 which I believe you nixed in offline communication but I can't remember the rationale.  Alternately, we could do the synchronization in `standalone-enc/main.c` directly, but that seems messier to me.).

gperciva:
Here's a graphical representation.  (aesctr is ~1600 MB/s, so it's off the top of this plot and is clearly not a bottleneck)

compilers

gperciva:
So far I haven't gotten dtrace working, but here's a flame graph from linux (inside virtualbox).  We can clearly see that encryption isn't the bottleneck, although I'm somewhat at a loss as to what to investigate next.  (Other than dtrace on freebsd, since you're more familiar with bsd networking & kernel)

proto-pipe
(github doesn't allow svg, so it's png)

cperciva:
It's entirely plausible that there's a big difference between the raw crypto and spiped.  Back when all the crypto was being done in software, spiped's overhead was small, but it's going to be far more significant now.

This is one of the reasons spiped exists -- as a test case for libcperciva networking code.

gperciva:
So does this benchmark approach look useful / should I clean it up and submit the preliminary PRs?

cperciva:
I'm not sure yet -- there's a lot of code here.  At this point I would be more inclined to profile the entire spiped process, and hold off on standalone tests until overall system profiling tells us where to look.

Is there a straightforward way for me to obtain the stunnel and spiped performance numbers you mentioned before so I can test that locally?

cperciva:
LGTM.  Any reason this is still marked as a draft?

gperciva:
There's a few questions I'm not certain about:

- I'm still not confident about testing with dtrace; this whole PR was kind-of an alternative to messing around with shell scripts (to set up a whole spiped testing infrastructure), and inserting dtrace into the whole thing.  Is it worth having it all as a self-contained binary?

  (I think it is, but I'm not confident about asserting that when I'm still not comfortable with dtrace)

- even if we do want this in the binary, should the `pipe_enc` functions be put into a separate file; like we did with `tests/pthread_create_blocking_np/timing.h` and `timing.c` in https://github.com/Tarsnap/spiped/pull/327/files#diff-b3cac23013bc0fe4c13672c58d9d81fc95ee864c2876e2b6e44a52a56fb6a562?

gperciva:
EDIT: nope, we don't want to try moving this code into a separate `.c` file, because that gets messy with the static consts such as `perfsizes[]`, and the `perftest_buffers()` framework.

I mean, it wouldn't be impossible to separate, but it ends up looking weird, and not worth it for this testing code.

gperciva:
I've re-discovered why I wanted this: it's an easy way of benchmarking what happens when we use kivaloo's `netbuf_read` (and eventually `netbuf_write`, not included in this plot).  29% decrease in time due to `netbuf_read`:

spiped-netbuf-read

(also, I've increased the x-axis rotation, so future plots will be more readable)

PR #317, opened 2021-07-03
Perftest improvement

gperciva:


gperciva:
Adds three capabilities to `perftest.h`:
- the `init_func()` can be `NULL`
- add a `clean_func()`, which can also be `NULL`
- adds a choice of cputime or wall-clock time.

I've been thinking about adding a `clean_func` for a while, but it seemed like over-engineering.  But I have that capability in enough "draft" branches that I think it's worth adding it now.  Ditto for adding the ability to use wall-clock time instead of cputime.

libcperciva/util/perftest.c
```diff
@@ -11,21 +11,23 @@
...
+ * the time and speed of processing each buffer size.  ${init_func} and
+ * ${clean_func} may be NULL.
  */
 int
 perftest_buffers(size_t nbytes, const size_t * sizes, size_t nsizes,
     size_t nbytes_warmup,
     int init_func(void * cookie, uint8_t * buf, size_t buflen),
     int func(void * cookie, uint8_t * buf, size_t buflen, size_t num_buffers),
-    void * cookie)
+    int clean_func(void * cookie), void * cookie)
```
cperciva, line 31: If we're going to add `clean_func` then we should probably call it in the error path, i.e. change a lot of the `goto err2` to `goto err3`?

gperciva:
Added as a REBASE commit.

cperciva:
Please rebase.

gperciva:
Rebased this, as well as https://github.com/Tarsnap/libcperciva/pull/409.

PR #320, opened 2021-08-09
tests/nc-server: add "averaging" to rate-limited echo server

gperciva:


gperciva:
This PR is lacking a `make Makefiles`, but it's such a small change (adding a dependency on `monoclock.h`) that I figured it's not worth a whole commit.  We'll get that the next time we run `make Makefiles` for some other change.

PR #321, opened 2021-08-11
tests/pushbits: add warning about EOF on stdin

gperciva:


gperciva:
I don't know if we want to include this warning or not.

This issue cost me a few hours when working on `pthread_create_blocking()`, because my test case was based on `echo_stdin_stdout()` and I thought that the thread's non-existence was a problem with the creation code.  It took me a while to discover that `read(STDIN_FILENO, ...)` was giving me an EOF when running in github actions but not on my own systems (virtualized or not).

gperciva:
Also, even if we want to include the warning, the final paragraph of the commit message is rather less formal than the previous paragraphs, so you might prefer that to be rewritten or omitted.  I figured it was useful info to include while you were evaluating this.

tests/pushbits/main.c
```diff
@@ -75,7 +75,13 @@ echo_stdin_stdout(void)
 		goto err0;
 	}
 
-	/* Wait a short while, then cancel the thread. */
+	/*
+	 * Wait a short while, then cancel the thread.  Note that if stdin
+	 * returned an EOF, the thread might have already stopped, and
+	 * certainly will not be open after a second.  This is particularly
```
cperciva, line 81: s/open/running/ ?

gperciva:
Rebased, ready for merge.

PR #327, opened 2021-10-27
pthread_create_blocking alternate api

gperciva:


lib/util/pthread_create_blocking.h
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * pthread_create_blocking(thread, attr, init_routine, run_routine,
+ *     cleanup_routine, arg):
+ * Run pthread_create() and block until the new ${thread} has started and
+ * ${init_routine} has finished.  ${init_routine} and ${cleanup_routine} may
+ * be NULL.  If ${init_routine} is not NULL and returns a non-zero value, do
+ * not attempt to run ${run_routine} and instead return that value.  When
+ * ${run_routine} finishes, make its returned value available via
+ * pthread_join().
```
cperciva, line 14: This comment tells me nothing about what `cleanup_routine` does except that it may be NULL.

lib/util/pthread_create_blocking.c
```diff
@@ -0,0 +1,279 @@
...
+{
+	struct wrapped_cookie * U;
+	int rc = -1;
+	int rc_init = 0;
+	int rc_sync = 0;
+	int initialized = 0;
+	pthread_mutex_t mutex;
+	pthread_cond_t cond;
+
+	/* Allocate our cookie and assign user-specific values. */
```
cperciva, line 181: s/assign user-specific values/store parameters/ ?

lib/util/pthread_create_blocking.c
```diff
@@ -0,0 +1,279 @@
...
+	}
+	if ((rc = pthread_cond_init(&cond, NULL)) != 0) {
+		warnp("pthread_cond_init: %s", strerror(rc));
+		goto err2;
+	}
+	U->mutex_p = &mutex;
+	U->cond_p = &cond;
+	U->initialized_p = &initialized;
+	U->rc_init_p = &rc_init;
+	U->rc_sync_p = &rc_sync;
```
cperciva, line 204: Why are you creating these variables on the stack and then passing pointers to them?  Wouldn't it be cleaner to just put them inside the structure?

lib/util/pthread_create_blocking.c
```diff
@@ -0,0 +1,279 @@
...
+ * pthread_create_blocking(thread, attr, init_routine, run_routine,
+ *     cleanup_routine, arg):
+ * Run pthread_create() and block until the new ${thread} has started and
+ * ${init_routine} has finished.  ${init_routine} and ${cleanup_routine} may
+ * be NULL.  If ${init_routine} is not NULL and returns a non-zero value, do
+ * not attempt to run ${run_routine} and instead return that value.  When
+ * ${run_routine} finishes, make its returned value available via
+ * pthread_join().
+ */
+int
```
cperciva, line 167: Can we make `pthread_create_blocking` have the same return values as `pthread_create`, i.e. return an error number?

Normally in cpercivaware we use -1 to indicate failure but it seems that behaving like `pthread_create` should take priority here.

lib/util/pthread_create_blocking.c
gperciva: I had a reason.  I'm not saying it was a good reason.  I wanted to check `rc_init` after `free(U)`.  Other things grew out of that.

Your suggestion does sound cleaner, so I'll adjust accordingly.

lib/util/pthread_create_blocking.c
gperciva: Do you mean "if `pthread_create` fails, return that error number", or "if any `pthread_*` fails, return that error number?

I now have it returning a `pthread_create` failure, but if something like `pthread_mutex_lock` fails, it prints a warning and returns the generic -1.

cperciva: I'd say if any error occurs.  You can't return a generic -1 for some errors but not others; the caller should be treating nonzero returns as errno values, and that would result in `Operation not permitted` being bogusly printed.

lib/util/pthread_create_blocking.c
```diff
@@ -0,0 +1,279 @@
...
+	/* 1 second later than the current time. */
+	if (monoclock_get(&tv)) {
+		warnp("monoclock_get");
+		goto err0;
+	}
+	tp.tv_sec = tv.tv_sec;
+	tp.tv_nsec = 1000 * tv.tv_usec;
+
+	/* Wait for up to 1 second. */
+	if ((rc = pthread_cond_timedwait(U->cond_p, U->mutex_p, &tp)) != 0) {
```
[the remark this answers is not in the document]
cperciva: I'm inclined to just say that if `pthread` functions are breaking, we're probably going to end up with deadlocks sooner or later so we might as well just accept it.  If you're really worried, mark the mutex as robust so that `pthread_cond_wait` will return if the spawned thread dies.

lib/util/pthread_create_blocking.c
gperciva: What should we do if

    int init_routine(void * cookie)

returns non-zero?

Currently, I'm returning that value from `pthread_create_blocking()`.  It seems useful to keep that functionality, but there could be confusion between the returned values being `errno` vs. being from the init.

The simplest solution I can think of is to suggest (in the API comment) that `init_routine` should return a negative value.  Then the calling code can treat rc < 0 as being from init, and rc > 0 as being errno.

gperciva: On second thought: why have an `init_routine` at all?  I mean, what's the use case for:

* not initializing stuff before calling `pthread_create_blocking()`, and
* not waiting until after `cleanup_routine()` has been set up.

... come to think of it, maybe we should move the `cleanup_routine` stuff away from `pthread_create_blocking.c` as well.  I mean, if I were designing an API from scratch then I'd rather have that as part of the "run a thread" thing, but given that `pthread_create()` doesn't have it, it's probably better to go for API compatibility since everybody [1] already knows what to expect from `pthread_create()`.

[1] everybody who's looking at the source code for cpercivaware, at least.

gperciva: Yeah, I'm nuking `init_routine` and `cleanup_routine`.  Simplifies a lot of things, and it allows a trivial diff when updating `pushbits.c` to use `_blocking()`.

gperciva:
Updated with REBASE commits.

> I'm inclined to just say that if pthread functions are breaking, we're probably going to end up with deadlocks sooner or later so we might as well just accept it.

I'm not certain that I follow.  The whole point is that if we find that pthread functions are breaking, we don't run the provided routine, and `pthead_create_blocking()` returns with an error.  In theory, the polling loop should discover that something's very wrong within 1 second, and quit.

Granted, I expect that in such a case, the calling code would just head to an `exit(1)`.

Commit 8f9e8ed removes that error-handling code.  I think it's fairly harmless and worth keeping, but up to you.

lib/util/pthread_create_blocking.c
```diff
@@ -0,0 +1,162 @@
...
+		 * Don't record any pthread_mutex_unlock() error here; the
+		 * error from pthread_create() is more important.
+		 */
+		pthread_mutex_unlock(&U->mutex);
+		goto err3;
+	}
+
+	/* Wait for the thread to have started, then unlock the mutex. */
+	while (!U->initialized) {
+		/* Wait up to 1 second. */
```
gperciva, line 125: Oops, forgot to update that comment.

I'll hold off on changing it, pending your response about reverting 8f9e8ed (i.e. keeping the "check if pthreads mutex & condition variables are broken").

lib/util/pthread_create_blocking.c
```diff
@@ -81,36 +76,6 @@ wrapped_thread(void * cookie)
...
-	struct timespec tp = {0};
-	struct timeval tv;
-	int rc;
-
-	/* 1 second later than the current time. */
-	if (monoclock_get(&tv)) {
-		rc = errno;
-		goto err0;
-	}
-	tp.tv_sec = tv.tv_sec;
```
gperciva: huh, I wonder how this worked at all?  I forgot to `+ 1` here.

gperciva:
I'm leaning towards reinstating the provided function being required to call

    init_func(init_cookie);

The current code in this PR does what it says: it blocks until ${thread} has started.  But that's different from blocking until ${start_routine} has started, and I can fairly reliably (maybe 50% of the time) have the `pthread_create_blocking` return before the ${start_thread} has begun:

```
0.00000000	1	main_start
0.00003700	1	thr_start
0.00004200	1	thr_stop
0.00003700	1	main_after_block
0.00007900	1	main_after_join
0.00000000	2	main_start
0.00008200	2	thr_start                           <---------- here
0.00008300	2	thr_stop
0.00007800	2	main_after_block             <---------- here
0.00009100	2	main_after_join
test_pthread_create_blocking: timing.c, 195
test_pthread_create_blocking: info2.thr_start is after info2.main_after_block
```

("<--- here" editorial marks)

Now, we could argue that ${start_routine} is very likely to start very soon after ${thread} has begun, so that's enough for `pthread_create_blocking()`.  But that seems like a weak argument to me.  If the caller can't rely on ${start_routine} to have begun, then why bother with `pthread_create_blocking()` at all?

cperciva:
I thought the point of this was just to make sure that the thread ID was actually valid when the create function returns?

gperciva:
... I can't believe that I forgot that.  Huh.

cperciva:
Why have a cleanup routine?  The only difference between doing it here and having it in `start_routine` is in the case where some very basic pthread routines break, and that's basically a situation of "the OS is broken and there's nothing we can do to recover" anyway.

cperciva:
... I mean, if `pthread_mutex_*` or `pthread_cond_*` aren't working, what's the odds that `pthread_cleanup_*` will work anyway?

gperciva:
I'm imagining the scenario of the caller doing:

```
if (pthread_create_blocking(thr, ...) != 0)
    goto err0;
pthread_cancel(thr);
```

If `pthread_cleanup` is part of `start_routine`, then it may or may not be set up, so we're left with indeterminately-cleaned-up resources.

(not that it really matters, but I figure that if we're going to name something after `pthread_create`, it should be as bullletproof as possible)

gperciva:
... on a related note, it occurred to me that `pthread_` is probably a reserved identifier.  (I haven't tracked down the exact reference yet.)

Should we call it something else like `create_thread_blocking`?  I can't think of any good alternate name right now, but I'm reluctant to use a technically-invalid name.

gperciva:
> should be as bullletproof as possible

Just to elaborate on this... you've mused about other groups potentially using some libcperciva files (I think it was parsenum.h in particular).  I'm thinking of `pthread_create_blocking` in a similar way, i.e. trying to make sure it works (as well as possible) in all situations, not just what we want for spiped in particular.

cperciva:
> I'm imagining the scenario of the caller doing:
> 
> ```
> if (pthread_create_blocking(thr, ...) != 0)
>     goto err0;
> pthread_cancel(thr);
> ```
> 
> If `pthread_cleanup` is part of `start_routine`, then it may or may not be set up, so we're left with indeterminately-cleaned-up resources.

Meh, if you call `pthread_cancel` it's up to you to ensure that you know what you're cancelling.  In particular, the thread should set up its cleanup handlers appropriately with respect to creating resources; when the thread is first created there's nothing to clean up yet, of course.

cperciva:
> ... on a related note, it occurred to me that `pthread_` is probably a reserved identifier. (I haven't tracked down the exact reference yet.)
> 
> Should we call it something else like `create_thread_blocking`? I can't think of any good alternate name right now, but I'm reluctant to use a technically-invalid name.

Meh.  Yes, it's reserved; but I'm not too concerned about this.  Maybe call it `pthread_create_blocking_np` though; the `_np` suffix is widely used to denote non-POSIX extensions to pthreads.

gperciva:
Ok, I'll remove the `cleanup` stuff again.  I forgot that `pthread_cancel` only takes effect at a cancellation point, so the provided function can safely ensure that it sets up its own cleanup (if it wants one) before .

Can I rebase this branch?  It's such a mess now, keeping REBASE commits doesn't make it easier to see what's happening.

gperciva:
> when the thread is first created there's nothing to clean up yet, of course.

As a quibble, that wouldn't allow a thread to take ownership of resources.

For example, in `spipe/pushbits.c` we have:

- main: `pushbits()` allocates `struct push * P`.
- thread: `workthread()` sets up `workthread_cleanup`, which does `free(P)`.

But that's ok with the current design, because the `${start_routine}` can set up its own cleanup function before any cancellation points.

cperciva:
Yes, feel free to clean up the history.  I'm only looking at the final version of the file right now anyway.

cperciva:
> > when the thread is first created there's nothing to clean up yet, of course.
> 
> As a quibble, that wouldn't allow a thread to take ownership of resources.

It could, but I agree that it's more work.  (e.g. by locking a shared mutex and setting a `thread_has_taken_ownership` variable.)

lib/util/pthread_create_blocking_np.c
```diff
@@ -0,0 +1,168 @@
...
+
+struct wrapped_cookie {
+	/* User-supplied. */
+	void * (*start_routine)(void *);
+	void * arg;
+
+	/* Synchronization. */
+	pthread_mutex_t mutex;
+	pthread_cond_t cond;
+	int initialized;	/* 1 if initialized. */
```
cperciva, line 17: Rather than "initialized" -- which I think is a reference to the no-longer-present initialization callback? -- how about just "running"?

gperciva:
Rebased branch, but included one MAYBE commit.

8 days ago, you said:
> I'm inclined to just say that if pthread functions are breaking, we're probably going to end up with deadlocks sooner or later so we might as well just accept it. If you're really worried, mark the mutex as robust so that pthread_cond_wait will return if the spawned thread dies.

I don't think that I understand.  As far as I can tell, as long as a breaking pthread function returns a non-zero value, we can at least ensure that `pthread_create_blocking_np()` returns non-zero.

I know that so far you've been looking at the overall "files changed", but please take a glance at this single commit:
071e164 MAYBE add polling check for signalling error

Without this, if there's a problem with `pthread_lock` or `pthread_cond_signal`, `pthread_create_blocking_np()` can block indefinitely because `U->cond` was never signalled.

With this commit, the worst case is that `pthread_create_blocking_np()` will return an error within a second of such a failure.

lib/util/pthread_create_blocking_np.c
```diff
@@ -0,0 +1,170 @@
...
+ */
+int
+pthread_create_blocking_np(pthread_t * restrict thread,
+    const pthread_attr_t * restrict attr,
+    void * (*start_routine)(void *), void * restrict arg)
+{
+	struct wrapped_cookie * U;
+	int rc;
+
+	/* Allocate our cookie and store parameters. */
```
cperciva, line 128: Worth a comment here mentioning that the C standard does not require that
variables with automatic storage duration are accessible by other threads;
POSIX (and thus pthreads) does provide that guarantee, but out of an abundance
of caution we take the option of using malloc in case this code ends up
running on a non-POSIX system.

lib/util/pthread_create_blocking_np.c
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	/* Wait for the thread to have started, then unlock the mutex. */
+	while (!U->running) {
+		/* Wait until signalled. */
+		if ((rc = pthread_cond_wait(&U->cond, &U->mutex)) != 0)
+			goto err5;
+
+		/* Quit if there was an error in the synchronization. */
+		if (U->rc_sync) {
```
cperciva, line 168: For consistency, `if ((rc = U->rc_sync) != 0))`

lib/util/pthread_create_blocking_np.c
```diff
@@ -0,0 +1,170 @@
...
+/* Routine which is executed by pthread_create(). */
+static void *
+wrapped_thread(void * cookie)
+{
+	struct wrapped_cookie * U = cookie;
+	void * (*start_routine)(void *);
+	void * arg;
+	int rc;
+
+	/* Copy the user-supplied parameters, because U will not be valid. */
```
cperciva, line 68: ... after we signal the main thread that we have started running.

lib/util/pthread_create_blocking_np.c
```diff
@@ -0,0 +1,170 @@
...
+	start_routine = U->start_routine;
+	arg = U->arg;
+
+	/* Indicate that the main thread can continue (which will free U). */
+	if ((rc = set_running(U)) != 0) {
```
cperciva, line 73: I would merge `set_running` into here.  I don't see any benefit to having it as a separate function.

lib/util/pthread_create_blocking_np.c
```diff
@@ -0,0 +1,170 @@
+#include <sys/time.h>
```
cperciva, line 1: ?

cperciva:
I don't like the MAYBE.  On the positive side, it would avoid a deadlock if... err... very low-level pthreads functions are broken?  But on the downside, it could introduce failures if a system is so heavily loaded that it takes more than a second before a new thread starts running.

lib/util/pthread_create_blocking_np.c
gperciva: I thought it would be easier to reason about?  I mean, `set_running` is the bog-standard way to signal a condition variable.  I figured that if that was a separate function, it would be easy to check, then the reader could progress to `wrapped_thread`, which would also be easier to understand due to the signalling and mutexes not being part of it.

I've merged it as you requested, but I put that as the final commit in this branch, in case you take a look at the merged code and reconsider.

gperciva:
Ok, I've reverted the MAYBE, and added REBASE commits for everything you noted.

gperciva:
> But on the downside, it could introduce failures if a system is so heavily loaded that it takes more than a second before a new thread starts running.

Wait, I don't understand this part.

If `pthread_cond_timedwait` in 071e164 ends with a ETIMEDOUT, we ignore the error and `cond_timedwait` returns with a 0 so there's no `goto err5`.  `U->rsync` is also still a 0, so there's also no `goto err5`.  `U->running` is still 0, so we start another 1-second timed wait.  We keep on looping -- checking if there's an error every second -- until either the thread has started, or we've found an error.

(I've already reverted the commit in this branch; I'm asking for general pthread knowledge.)

gperciva:
Added a MAYBE comment in c5900e43 to explain why we're not using polling.

cperciva:
> > But on the downside, it could introduce failures if a system is so heavily loaded that it takes more than a second before a new thread starts running.
> 
> Wait, I don't understand this part.
> 
> If `pthread_cond_timedwait` in [071e164](https://github.com/Tarsnap/spiped/commit/071e164f7c76c2c1a20d9e367dda7ef20e98fe0c) ends with a ETIMEDOUT, we ignore the error and `cond_timedwait` returns with a 0 so there's no `goto err5`. `U->rsync` is also still a 0, so there's also no `goto err5`. `U->running` is still 0, so we start another 1-second timed wait. We keep on looping -- checking if there's an error every second -- until either the thread has started, or we've found an error.
> 
> (I've already reverted the commit in this branch; I'm asking for general pthread knowledge.)

Never mind, I was misreading the code.  Yeah, not dangerous; just useless.  In fact, even more useless than I realized at first, since pthread_cond_timedwait isn't guaranteed to return after the specified time; rather, after the specified time *it tries to reacquire the mutex and blocks until that succeeds*.  So the case of the pthread library being broken and is likely to result in this polling not actually polling anyway.

cperciva:
I think the MAYBE comment will add more confusion than it removes, at least among people familiar with pthreads.  But the rest looks good; please rebase.

Maybe worth s/main thread/parent thread/ everywhere?  Since conceivably any thread could spawn a new one.  Up to you; both options are clear, I think.

gperciva:
Rebased with "parent thread" instead of "main thread".  Potentially ready for merge.

gperciva:
Wait, I forgot to update "main -> parent" in the test file `timing.c`.

gperciva:
Fixed, potentially ready for merge.

PR #331, opened 2021-12-05
proto_pipe.c: simplify network calls

gperciva:


cperciva:
> proto_pipe.c: use P->inbuf, P->outbuf

I would be tempted to keep the existing dbuf/ebuf buffers and add new inbuf/outbuf fields which are pointers to the appropriate buffers.  Would that cause problems with #332?  I can live with your approach if it simplifies later work.

gperciva:
After #332, P->inbuf is removed, because `netbuf_read.c` stores the actual read buffer and just gives us a pointer to the relevant portion via `netbuf_read_peek()`.

I'll try sketching a different PR which keeps P->ebuf and P->dbuf in case the git history looks better, but unless I've drastically misunderstood something about netbuf_read, we wouldn't want to keep them after switching.

(I'm not certain what's happening with netbuf_write, because so far all of my attempts to use it resulted in it being slower.  But I'm still investigating.)

gperciva:
> After #332, P->inbuf is removed, because netbuf_read.c stores the actual read buffer and just gives us a pointer to the relevant portion via netbuf_read_peek().

To elaborate: once we're using `netbuf_read`, we only need `uint8_t * inbuf` in `callback_pipe_read()`, so I've made that a local function variable.

gperciva:
Now that I see it, I agree that this makes the history look nicer.

gperciva:
Rebased, because I wanted to fix the commit message.

As a future note: I still think that we should remove `P->ebuf` and `->dbuf` before switching to `netbuf_read()`, but that change should be in a separate commit with a longer explanatory message.  It doesn't need to happen in this "simplify network calls" PR.

cperciva:
> (I'm not certain what's happening with netbuf_write, because so far all of my attempts to use it resulted in it being slower. But I'm still investigating.)

Does it improve if you bump WBUFLEN in netbuf_write.c?

gperciva:
> Does it improve if you bump WBUFLEN in netbuf_write.c?

No, but I'm extremely suspicious about some of the other things I did in that branch.  But that doesn't impact netbuf_read.

PR #335, opened 2021-12-09
build: run `make Makefiles`

gperciva:


gperciva:
This could have been part of #332 but I wanted that PR to focus on code changes.

I assumed that I'd find some tiny build things to fix and wrap those into this PR.  But I haven't found anything else to do, and we need to do this before `make publish`.

PR #339, opened 2021-12-11
Fix build on arm32

gperciva:


gperciva:
re the modification to the CHANGELOG: as mentioned in https://github.com/Tarsnap/libcperciva/pull/425#issuecomment-991446804, the sha256 improvement should still work on an arm32 chip that has the cryptographic extensions.  But it's easier to phrase it as "aarch64", and if somebody on such a chip notices that sha256 is much faster, I doubt that they'll complain.

PR #340, opened 2021-12-11
build: do tests-clean as part of regular clean

gperciva:


gperciva:
While looking at this, it occurred to me that we're not using `.PHONY` everywhere that we should.  But I'm going to hold off on submitting that change until after 1.6.2.

PR #341, opened 2021-12-14
Allow ENOPROTOOPT for SO_REUSEADDR

gperciva:


gperciva:
Technically, a POSIX-compatible system is not required to allow SO_REUSEADDR for AF_UNIX sockets.  GNU/Hurd does not allow that, and AFAIK it is the only operating system to not do so.

My recommendation is to consider merging this after 1.6.2, because there's other commits which are required to work around bugs in GNU/Hurd.  (Based on my reading of the POSIX spec, this particular item is not a bug in GNU/Hurd.)

cperciva:
We should just allow ENOPROTOOPT for any socket types.  While AF_UNIX is the only case where it's known to fail, there's no guarantee that it can't happen in other cases; and we don't need to worry about an undetected failure since we would get a failure on bind anyway.

cperciva:
LGTM.  Remove the draft flag?

gperciva:
Did you want this merged before 1.6.2?  I was suggesting that we wait until after, but it's your call.

PR #343, opened 2022-02-04
Valgrind

gperciva:


gperciva:
Main updates:

- add a `| head -n1` after a grep in the valgrind test suite, so that if there's multiple `pl_*` functions it takes the first one, instead of breaking the whole suite.
- make the `getaddrinfo()` valgrind check ignore an `EAI_AGAIN`, which would almost certainly arise due to being offline.

PR #344, opened 2022-02-05
perftests/standalone-enc: add dtrace flamegraph script

gperciva:


gperciva:
This assumes that one has `doas(1)` set up to allow the current user to run `/usr/sbin/dtrace`.

Sample output:

bench

cperciva:
Can you make the use of `doas` conditional on `id -u`?  Check the uid, set DOAS, then invoke ${DOAS} instead of a hard-coded path.

That way this can be run as root on systems without (appropriately configured) doas.

perftests/standalone-enc/freebsd_flamegraph.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+outfilename="bench"
+
+# Assume that the user has doas(1) set up to allow /usr/sbin/dtrace.
+doas /usr/sbin/dtrace					\
+	-x ustackframes=100				\
+	-n 'profile-97 { @[ustack()] = count(); } tick-60s { exit(0); }' \
```
cperciva, line 8: Why is the `tick-60s` needed here?  It should exit when the test completes...?

cperciva:
We only care about stacks which include `pipe_enc_thread` here, right?  Can you stick a grep after stackcollapse.pl to filter out the rest?

gperciva:
> Can you stick a grep after stackcollapse.pl to filter out the rest?

Sure, but I thought the whole point of the interactive svg was that you could click on the part that you wanted to see more details about?

I didn't want to assume that the stuff in the non-`pipe_enc_thread` threads (`noeintr_write` and `read`) were irrelevant.  I mean, if the perftest is doing what it's supposed to, then of course they're not relevant... but I wouldn't want to assume that it's perfect.

Hmm, probably best to make this an option to the shell script.  Grep it out by default, but allow the user to keep everything if they want.

perftests/standalone-enc/freebsd_flamegraph.sh
```diff
@@ -0,0 +1,53 @@
...
+# To see all profile data:
+#       ./freebsd_flamegraph.sh --all
+
+set -e -o noclobber -o nounset
+
+outfilename="bench"
+
+# Should we keep all the profile data?
+KEEP_ALL=0
+if [ "$#" -eq "1" ]; then
```
cperciva, line 15: Not sure I understand why this is necessary.

perftests/standalone-enc/freebsd_flamegraph.sh
```diff
@@ -0,0 +1,53 @@
...
+	DTRACE_CMD="doas /usr/sbin/dtrace"
+fi
+
+# Clear previous data
+rm -f "${outfilename}.stacks" "${outfilename}.folded" "${outfilename}.svg"
+
+# Get profile data and consolidate it.
+${DTRACE_CMD}						\
+	-x ustackframes=100				\
+	-n 'profile-97 { @[ustack()] = count(); }'	\
```
cperciva, line 36: I would use `profile-997` rather than `profile-97` in order to get more data.  Also perhaps add an option to `test_standalone_enc` to increase the amount of data it handles; the current 100 MB doesn't allow for many data points to be gathered.

You should also ask dtrace to filter the profiling hits: We only want to record when execname is "test_standalone_enc" and arg1 is nonzero.  Otherwise we'll gather data about other processes on the system and time spent in the kernel.  Add `/execname == "test_standalone_enc" && arg1/` after `profile-N`.

gperciva: Doesn't the `-c` take care of this?  I'm not seeing info about other processes?  Or maybe that's just because I'm only running vim at the same time as profiling, and that has minimal resource usage?

Since it's only counting `ustack()` I thought it wouldn't be picking up kernel stuff.

But ok, I'll do the `/execname` thing.

gperciva: Granted, the man-page makes no claims about `-c` only gathering info about the specified command, so ok.

cperciva: The long "tail" on the left of your SVG is dtrace reporting all the times the probe fired and the kernel was running the idle thread.

gperciva:
Updates.  Now the binary can take an optional value to multiply the number of bytes to process, while the script is hard-coded to make that 10 (so it does 1 GB instead of 100 MB).  Takes 7.8 seconds on my server.

Do you want the script to also allow an optional multiplier?

perftests/standalone-enc/main.c
```diff
@@ -580,18 +580,27 @@ int
...
 	int desired_test;
+	size_t multiplier;
 
 	WARNP_INIT;
 
 	/* Parse command line. */
-	if (argc != 2) {
-		fprintf(stderr, "usage: test_standalone_enc NUM\n");
+	if ((argc < 2) || (argc > 3)) {
+		fprintf(stderr, "usage: test_standalone_enc NUM MULT\n");
```
cperciva, line 589: `[MULT]`

cperciva:
Looks good, just one minor nit.  Please rebase.

gperciva:
Rebased, ready for merge.

PR #345, opened 2022-02-19
standalone-enc cleanup

gperciva:


cperciva:
BTW I can't remember if I've mentioned this before, but for static functions I'm generally ok with `/* Single line comment */` rather than a full multi-line comment.

gperciva:
Yeah, that's included in the `STYLE`.  But I wanted to split that file into separate perftests (in the new #346(, so having a full comment made it easier to copy&paste.

We also sometimes have multi-line comments for static functions, so I figured there was no harm in introducing them now even if you don't like the other PR.

cperciva:
Ah, now I see the context.  Yeah, splitting these into different files is good (and requires the multi-line comments anyway).

PR #346, opened 2022-03-10
Split standalone-enc/main.c into separate files

gperciva:


gperciva:
`main.c` was 661 lines long, and it was getting annoying to see exactly what each perftest did.  This will simplify adding perftests in the future: we can start by copying one file, then modifying it.  Also, we can easily diff the code for two different perftests, if we're investigating something.

PR #348, opened 2022-04-11
Libcperciva import queue.h

gperciva:


COPYRIGHT
```diff
@@ -1,3 +1,6 @@
+The code in the "libcperciva/external" directory is distributed
```
gperciva, line 1: I'm not completely certain where you want this copyright notice to be.

gperciva: Hmm, tarsnap has a top-level COPYRIGHT and a libcperciva/COPYRIGHT.  But spiped only has a top-level COPYRIGHT.

PR #350, opened 2022-04-12
proto_pipe.c: check for an error in shutdown()

gperciva:


gperciva:
Coverity-scan warned us that this is the only instance of `shutdown()` for which we don't check the return value.  I can think of 3 ways of addressing it:

- My first instinct was just to slap a `warnp("shutdown")` there, but then I noticed that `proto_pipe.c` doesn't have any other warnp.
- Another option is just to ignore it with `(void)`, since this is in the `eof:` section anyway.
- Finally, we could do what this PR does: report an error to the upstream code, and let it investigate `errno` (or not).

lib/proto/proto_pipe.c
```diff
@@ -162,11 +163,14 @@ callback_pipe_read(void * cookie, int status)
 	return ((P->callback)(P->cookie));
 
 eof:
-	/* We aren't going to write any more. */
-	shutdown(P->s_out, SHUT_WR);
-
-	/* Record that we have reached EOF. */
-	*(P->status) = 0;
+	/* We aren't going to write any more; ignore any ENOTSOCK. */
+	if (shutdown(P->s_out, SHUT_WR) && (errno != ENOTSOCK)) {
```
cperciva, line 167: Hmm, I'm not sure that ENOTSOCK is valid here.  These are descriptors we're calling `select()` on, so they had better be sockets.

For that matter, if we get EBADF or EINVAL those should be fatal errors -- it means something went wrong, not just that the connection is dying.  And I *think* ENOTCONN is wrong too but I'm not 100% certain.  (If a connection is being killed e.g. by a RST arriving, I think the socket still counts as being "connected" even though the TCP connection it applies to is dead?)

My inclination is to make this a `warnp` / `goto err0;` situation, possibly with special handling for ENOTCONN.

gperciva:
Hmm.  It might be a while until I know enough about networking to have an opinion about ENOTCONN in this context.

From a general software view, this version does not use `warnp` because nothing else in this file does -- it seems that you preferred to do `go err0` directly.

cperciva:
Looks like I just didn't have anywhere which needed `warnp`.  The error paths in this file were all non-fatal (e.g. a connection being closed) or functions which fail due to running out of memory.  (Note also that this file mostly just calls library functions, not syscalls).

But it shutdown fails for a should-not-happen reason we should warn.

gperciva:
Added `warnp`, ready for merge.

cperciva:
Ok, I *think* this is ok; it looks to me like ENOTCONN should not (non-erroneously) happen here, even if we get RST.

If users start reporting ENOTCONN then we should probably treat it as a non-error.

Issue #351, opened 2022-04-13
local interface for outgoing connections


nerijus:
I need to use non default interface for outgoing connections. telnet has `-b` option for it:
```
$ telnet 10.10.10.63 8080
Trying 10.10.10.63...

telnet: connect to address 10.10.10.63: Connection timed out

$ telnet -b 10.15.0.2 10.10.10.63 8080
Trying 10.10.10.63...

Connected to 10.10.10.63.
```
Is it possible when using spiped?

cperciva:
There is no support for this yet, but it should be easy to add.  Are you interested in writing a patch?

nerijus:
I could try if you guide me where in the code (I took a look at the code already).

cperciva:
Ok, there's unfortunately a lot of plumbing to get the bind address where it needs to go, but it's reasonably straightforward despite being annoying how many places need to be touched.

You'll want to add an option (`-b` sounds good to me) to the getopt loop in spiped's `main.c`.  Pass it to `sock_resolve` and then the `struct sock_addr *` needs to be:

1. Passed as a new parameter to `dispatch_accept`;
2. Stored as a new field in `struct accept_state`;
3. Passed as a new parameter to `proto_conn_create`;
4. Passed as a parameter to `network_connect_bind` which is the same as `network_connect` except with an added argument;
5. Stored as a new field in `struct connect_cookie`;
6. Passed as a parameter to `sock_connect_bind_nb` which is the same as `sock_connect_nb` except with an added argument;

and finally that function can call `bind` after creating a socket.

Once you've done this, you can also add it to spipe with far less work: Just add the option to the getopt loop and pass it to `proto_conn_create`.

nerijus:
Thank you for the guide!
1.  As `-b` is optional, can I pass `NULL` as `struct sock_addr ** sas_b` (a new parameter) to `dispatch_accept` when `-b` is not used?

cperciva:
I would pass `NULL` to indicate that we don't want to bind to any specific address.

Also, pass in a `struct sock_addr *`, not a `**` -- you'll get back a `**` (which is a list of addresses) from `sock_resolve` but we can only bind to one address so just ignore any extras (possibly with a warning message).

nerijus:
Thank you for easy to follow and complete guide. I created PR https://github.com/Tarsnap/spiped/pull/353
[...]

PR #353, opened 2022-04-26
Add option "-b" which allows to bind to non default address for outgoing connections

nerijus:


libcperciva/network/network_connect.c
```diff
@@ -182,6 +188,24 @@ network_connect(struct sock_addr * const * sas,
...
+ * ${callback}(${cookie}, -1).  Return  a cookie which can be passed to
+ * network_connect_cancel() in order to cancel the connection attempt.
+ */
+void *
+network_connect_bind(struct sock_addr * const * sas, struct sock_addr const * sa_b,
+    int (* callback)(void *, int), void * cookie)
+{
+
+	/* Let network_connect_timeo handle this. */
+	return (network_connect_timeo(sas, NULL, callback, cookie));
```
cperciva, line 206: Looks like you're not actually using `sa_b` here?

spipe/main.c
```diff
@@ -113,7 +113,7 @@ usage(void)
 {
 
 	fprintf(stderr,
-	    "usage: spipe -t <target socket> -k <key file>"
+	    "usage: spipe -t <target socket> -k <key file> -b <bind address>"
```
cperciva, line 116: Since this is an optional argument, it needs to be in `[square brackets]`.

spipe/main.c
```diff
@@ -222,20 +229,28 @@ main(int argc, char * argv[])
 	ET.connection_error = 0;
 	ET.stopped = 0;
 
+	/* Resolve bind address. */
+	if (opt_b) {
+	    if ((sa_b = sock_resolve(opt_b)[0]) == NULL) {
```
cperciva, line 234: Need to store the complete list returned by `sock_resolve` in a variable.

Check that it's not NULL (in case `sock_resolve` failed).

And free the entire list later (not just the one address we end up using).

spipe/main.c
```diff
@@ -222,20 +229,28 @@ main(int argc, char * argv[])
...
+		warn0("No address found for %s", opt_b);
+		goto err1;
+	    }
+	}
+
 	/* Resolve target address. */
 	if ((sas_t = sock_resolve(opt_t)) == NULL) {
 		warnp("Error resolving socket address: %s", opt_t);
 		goto err0;
```
cperciva, line 243: This will become `goto err1`.

spipe/main.c
```diff
@@ -309,26 +324,29 @@ main(int argc, char * argv[])
...
 	close(s[0]);
-err2:
+err3:
 	proto_crypt_secret_free(K);
-err1:
+err2:
 	sock_addr_freelist(sas_t);
+err1:
+	if (opt_b)
+		sock_addr_free(sa_b);
```
cperciva, line 349: If we set `sas_b = NULL` by default, this can just be `sock_addr_freelist(sas_b)` (since `free(NULL)` is safe).

spiped/dispatch.c
```diff
@@ -301,6 +304,8 @@ dispatch_shutdown(void * dispatch_cookie)
 		events_timer_cancel(A->dnstimer_cookie);
 	if (A->T != NULL)
 		dnsthread_kill(A->T);
+	if (A->sa_b != NULL)
+		sock_addr_free(A->sa_b);
```
cperciva, line 308: Don't want to free this here; we free it in `main`.

cperciva: (Or rather, we *should* free it in `main`.  Looks like we're not doing that yet though.)

libcperciva/network/network.h
```diff
@@ -39,6 +39,18 @@ void network_accept_cancel(void *);
 void * network_connect(struct sock_addr * const *,
     int (*)(void *, int), void *);
 
+/**
+ * network_connect_bind(sas, sa_b, callback, cookie):
+ * Iterate through the addresses in ${sas}, attempting to create and connect
+ * a non-blocking socket after binding to ${sa_b}.  Once connected, invoke
+ * ${callback}(${cookie}, s) where s is the connected socket; upon fatal error
+ * or if there are no addresses remaining to attempt, invoke
+ * ${callback}(${cookie}, -1).  Return  a cookie which can be passed to
```
gperciva, line 48: There's two spaces between "Return" and "a".

libcperciva/network/network_connect.c
nerijus: Fixed.

spipe/main.c
nerijus: Oops, fixed.

cperciva:
Thanks for all the updates; I'll try to look at them tomorrow.  @gperciva might be able to look at them more before I find time.

libcperciva/network/network_connect.c
```diff
@@ -179,16 +185,34 @@ network_connect(struct sock_addr * const * sas,
...
+ * ${callback}(${cookie}, -1).  Return a cookie which can be passed to
+ * network_connect_cancel() in order to cancel the connection attempt.
+ */
+void *
+network_connect_bind(struct sock_addr * const * sas, struct sock_addr const * sa_b,
+    int (* callback)(void *, int), void * cookie)
+{
+
+	/* Let network_connect_timeo handle this. */
+	return (network_connect_timeo(sas, sa_b, NULL, callback, cookie));
```
gperciva, line 206: clang nitpick:

```
../libcperciva/network/network_connect.c:206:37: error: incompatible pointer types passing 'const struct sock_addr *' to parameter of type 'struct sock_addr *const *' [-Werror,-Wincompatible-pointer-types]
        return (network_connect_timeo(sas, sa_b, NULL, callback, cookie));
                                           ^~~~
../libcperciva/network/network.h:59:84: note: passing argument to parameter here
void * network_connect_timeo(struct sock_addr * const *, struct sock_addr * const *,
                                                                                   ^
```

nerijus: Oops, fixed.

lib/proto/proto_conn.c
```diff
@@ -191,23 +191,24 @@ proto_conn_drop(void * conn_cookie, int reason)
...
+ * ${nokeepalive} is zero.  Drop the connection if the handshake or
  * connecting to the target takes more than ${timeo} seconds.  When the
  * connection is dropped, invoke ${callback_dead}(${cookie}).  Free ${sas}
  * once it is no longer needed.  Return a cookie which can be passed to
  * proto_conn_drop().  If there is a connection error after this
  * function returns, close ${s}.
  */
 void *
-proto_conn_create(int s, struct sock_addr ** sas, int decr, int nopfs,
+proto_conn_create(int s, struct sock_addr ** sas, struct sock_addr * sa_b, int decr, int nopfs,
```
gperciva, line 211: Looks like sa_b is not actually used in proto_conn_create?

libcperciva/util/sock.c
```diff
@@ -432,6 +432,48 @@ sock_connect_nb(const struct sock_addr * sa)
...
+sock_connect_bind_nb(const struct sock_addr * sa, const struct sock_addr * sa_b)
+{
+	int s;
+
+	/* Create a socket. */
+	if ((s = socket(sa->ai_family, sa->ai_socktype, 0)) == -1)
+		goto err0;
+
+	/* Bind the socket to sa_b. */
+	if ((bind(s, sa_b->name, sa_b->namelen)) == -1)
```
gperciva, line 451: Missing brackets!

This always does `goto err1`, regardless of whether there's an error or not.

libcperciva/util/sock.c
```diff
@@ -432,6 +432,48 @@ sock_connect_nb(const struct sock_addr * sa)
...
+		warnp("Error binding socket");
+		goto err1;
+
```
gperciva, line 454: Missing brackets!

This always does goto err1, regardless of whether there's an error or not.

(I previously highlighted the top of this problematic section, but github shows 4 lines above the indicated line, so that didn't show the problem as well as highlighting this line.)

spiped/main.c
```diff
@@ -274,50 +281,58 @@ main(int argc, char * argv[])
 			warnp_syslog(1);
 	}
 
+	/* Resolve bind address. */
+	if (opt_b) {
+	    if ((sa_b = sock_resolve(opt_b)[0]) == NULL) {
+		warn0("No address found for %s", opt_b);
+		goto err2;
+	    }
+	}
```
TimWolla, line 290: Indentation mixes tabs and spaces here.

libcperciva/network/network_connect.c
```diff
@@ -118,8 +119,13 @@ tryconnect(struct connect_cookie * C)
...
 	for (; C->sas[0] != NULL; C->sas++) {
 		/* Can we try to connect to this address? */
-		if ((C->s = sock_connect_nb(C->sas[0])) != -1)
+		if (C->sa_b != NULL) {
+		    if ((C->s = sock_connect_bind_nb(C->sas[0], C->sa_b)) != -1)
 			break;
+		} else {
+		    if ((C->s = sock_connect_nb(C->sas[0])) != -1)
+			break;
+		}
```
TimWolla, line 128: Indentation mixes tabs and spaces here.

libcperciva/util/sock.c
nerijus: Oops, I took it from other part of the source and missed brackets, sorry. Fixed.

spiped/main.c
gperciva: Right.

I also wonder whether it should be
```
if (opt_b && ((sa_b = sock_resolve(opt_b)[0]) == NULL)) {
```

My preference would be to separate the `(opt_b)` from the `sock_resolve()`, but I suspect that Colin might prefer it all on one line.

(@nerijus: I wouldn't bother changing this part until Colni comments)

spipe/main.c
nerijus: Done, for spiped too.

spipe/main.c
nerijus: Done for both spipe and spiped.

spiped/dispatch.c
nerijus: Fixed.

spiped/main.c
nerijus: Rewrote it.

libcperciva/network/network_connect.c
nerijus: Fixed.

lib/proto/proto_conn.c
nerijus: Sorry, forgot - fixed.

nerijus:
There is a warning:
```
../libcperciva/network/network_connect.c: In function ‘network_connect_timeo’:
../libcperciva/network/network_connect.c:227:10: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  C->sa_b = sa_b;
          ^
```
Should I remove `const`?

gperciva:
In https://github.com/Tarsnap/spiped/pull/354/commits/f439f1ae4c1720e52f385d336c173d1d1b054fb6, I added a `const` to the definition of `sa_b` inside the `struct`, but I haven't looked at this PR in sufficient detail to judge if that was the correct fix or not.

I'll note that `struct sock_addr * const * sas` contains a `const`, so I'd be surprised if `sa_b` didn't have one somewhere.

lib/proto/proto_conn.c
cperciva: This should be const.

lib/proto/proto_conn.h
```diff
@@ -15,22 +15,23 @@ enum {
...
+ * layer keep-alives (if applicable) on both sockets if and  only if
+ * ${nokeepalive} is zero.  Drop the connection if the handshake or
  * connecting to the target takes more than ${timeo} seconds.  When the
  * connection is dropped, invoke ${callback_dead}(${cookie}).  Free ${sas}
  * once it is no longer needed.  Return a cookie which can be passed to
  * proto_conn_drop().  If there is a connection error after this
  * function returns, close ${s}.
  */
-void * proto_conn_create(int, struct sock_addr **, int, int, int, int,
+void * proto_conn_create(int, struct sock_addr **, struct sock_addr *, int, int, int, int,
```
cperciva, line 34: const

libcperciva/network/network.h
```diff
@@ -40,13 +40,25 @@ void * network_connect(struct sock_addr * const *,
...
+ * network_connect_bind(sas, sa_b, callback, cookie):
+ * Iterate through the addresses in ${sas}, attempting to create and connect
+ * a non-blocking socket after binding to ${sa_b}.  Once connected, invoke
+ * ${callback}(${cookie}, s) where s is the connected socket; upon fatal error
+ * or if there are no addresses remaining to attempt, invoke
+ * ${callback}(${cookie}, -1).  Return a cookie which can be passed to
+ * network_connect_cancel() in order to cancel the connection attempt.
  */
-void * network_connect_timeo(struct sock_addr * const *, const struct timeval *,
+void * network_connect_bind(struct sock_addr * const *, struct sock_addr const *,
```
cperciva, line 51: I normally spell this as `const struct sock_addr *` rather than `struct sock_addr const *`.  (The declarations are equivalent so this is just a style issue.)

libcperciva/network/network.h
```diff
@@ -40,13 +40,25 @@ void * network_connect(struct sock_addr * const *,
...
     int (*)(void *, int), void *);
 
+/**
+ * network_connect_timeo(sas, sa_b, timeo, callback, cookie):
```
cperciva, line 55: Can we keep `network_connect_timeo` unchanged (i.e. don't add a bind address to it)?

libcperciva/network/network_connect.c
```diff
@@ -15,6 +15,7 @@ struct connect_cookie {
 	int (* callback)(void *, int);
 	void * cookie;
 	struct sock_addr * const * sas;
+	struct sock_addr * sa_b;
```
cperciva, line 18: const

libcperciva/network/network_connect.c
```diff
@@ -179,16 +185,34 @@ network_connect(struct sock_addr * const * sas,
...
+}
+
+/**
+ * network_connect_timeo(sas, sa_b, timeo, callback, cookie):
+ * Behave as network_connect_bind(), but wait a duration of at most ${timeo} for
  * each address which is being attempted.
  */
 void *
-network_connect_timeo(struct sock_addr * const * sas,
+network_connect_timeo(struct sock_addr * const * sas, struct sock_addr const * sa_b,
```
cperciva, line 215: Oh, I see now why you were adding functionality to `network_connect_timeo`.  Let's do this differently -- shunt the functionality to an internal `static void * network_connect_internal`, then have `network_connect`, `network_connect_bind`, `network_connect_timeo` all use that.

spiped/dispatch.c
```diff
@@ -18,6 +18,7 @@ struct accept_state {
 	int s;
 	const char * tgt;
 	struct sock_addr ** sas;
+	struct sock_addr * sa_b;
```
cperciva, line 21: const

spiped/dispatch.c
```diff
@@ -145,6 +146,7 @@ callback_gotconn(void * cookie, int s)
 {
 	struct accept_state * A = cookie;
 	struct sock_addr ** sas;
+	struct sock_addr * sa_b;
```
cperciva, line 149: Not needed

spiped/dispatch.c
```diff
@@ -169,7 +171,7 @@ callback_gotconn(void * cookie, int s)
 	node_new->A = A;
 
 	/* Create a new connection. */
-	if ((node_new->conn_cookie = proto_conn_create(s, sas, A->decr,
+	if ((node_new->conn_cookie = proto_conn_create(s, sas, sa_b, A->decr,
```
cperciva, line 174: `A->sa_b`

spiped/dispatch.c
```diff
@@ -199,25 +201,25 @@ callback_gotconn(void * cookie, int s)
...
+ * (if applicable) if and only if ${nokeepalive} is zero.  Drop connections if the
+ * handshake or connecting to the target takes more than ${timeo} seconds.  If
  * dispatch_request_shutdown() is called then ${conndone} is set to a non-zero
  * value as soon as there are no active connections.  Return a cookie which can
  * be passed to dispatch_shutdown() and dispatch_request_shutdown().
  */
 void *
 dispatch_accept(int s, const char * tgt, double rtime, struct sock_addr ** sas,
-    int decr, int nopfs, int requirepfs, int nokeepalive,
+    struct sock_addr * sa_b, int decr, int nopfs, int requirepfs, int nokeepalive,
```
cperciva, line 222: const

spiped/main.c
```diff
@@ -274,50 +282,65 @@ main(int argc, char * argv[])
 			warnp_syslog(1);
 	}
 
+	/* Resolve bind address. */
+	if (opt_b) {
+		while ((sas_b = sock_resolve(opt_b)) == NULL) {
+			if (!opt_D) {
+				warnp("Error resolving socket address: %s", opt_b);
+				goto err2;
```
cperciva, line 290: err1

cperciva:
Need to add `sock_addr_freelist(sas_b);` in the non-error exit path of the two `main` functions.

spiped/dispatch.c
nerijus: Fixed.

spiped/dispatch.c
nerijus: Fixed.

libcperciva/network/network_connect.c
nerijus: Fixed.

libcperciva/network/network.h
nerijus: Fixed.

lib/proto/proto_conn.h
nerijus: Fixed.

lib/proto/proto_conn.c
nerijus: Fixed.

spiped/dispatch.c
nerijus: Fixed.

spiped/dispatch.c
nerijus: Fixed.

spiped/main.c
nerijus: Fixed.

nerijus:
> `sock_addr_freelist(sas_b);` in the non-error exit path of the two `main` functions
Done.

libcperciva/network/network_connect.c
nerijus: Done, could you please check if I understood you correctly?

nerijus:
I've rebased my PR on the branch `libcperciva-connect-bind` (PR #354).

lib/proto/proto_conn.c
```diff
@@ -191,23 +191,24 @@ proto_conn_drop(void * conn_cookie, int reason)
...
+ *     timeo, callback_dead, cookie):
  * Create a connection with one end at ${s} and the other end connecting to
- * the target addresses ${sas}.  If ${decr} is 0, encrypt the outgoing data;
- * if ${decr} is nonzero, decrypt the incoming data.  If ${nopfs} is non-zero,
- * don't use perfect forward secrecy.  If ${requirepfs} is non-zero, drop
- * the connection if the other end tries to disable perfect forward secrecy.
- * Enable transport layer keep-alives (if applicable) on both sockets if and
- * only if ${nokeepalive} is zero.  Drop the connection if the handshake or
+ * the target addresses ${sas}.  Optionally bind to address ${sa_b}.  If
+ * ${decr} is 0, encrypt the outgoing data;  * if ${decr} is nonzero, decrypt
```
gperciva, line 198: Error in vim's reflow: it kept the initial " * " in the rewrapped text.  (That happens to me often.)

lib/proto/proto_conn.c
```diff
@@ -191,23 +191,24 @@ proto_conn_drop(void * conn_cookie, int reason)
...
+ * proto_conn_create(s, sas, sa_b, decr, nopfs, requirepfs, nokeepalive, K,
+ *     timeo, callback_dead, cookie):
  * Create a connection with one end at ${s} and the other end connecting to
- * the target addresses ${sas}.  If ${decr} is 0, encrypt the outgoing data;
- * if ${decr} is nonzero, decrypt the incoming data.  If ${nopfs} is non-zero,
- * don't use perfect forward secrecy.  If ${requirepfs} is non-zero, drop
- * the connection if the other end tries to disable perfect forward secrecy.
- * Enable transport layer keep-alives (if applicable) on both sockets if and
- * only if ${nokeepalive} is zero.  Drop the connection if the handshake or
+ * the target addresses ${sas}.  Optionally bind to address ${sa_b}.  If
```
gperciva, line 197: I'd suggest: Bind outgoing address to ${sa_b} if it is not NULL.

lib/proto/proto_conn.c
```diff
@@ -241,9 +242,15 @@ proto_conn_create(int s, struct sock_addr ** sas, int decr, int nopfs,
...
-	if ((C->connect_cookie =
-	    network_connect(C->sas, callback_connect_done, C)) == NULL)
-		goto err2;
+	if (sa_b == NULL) {
+		if ((C->connect_cookie =
+		    network_connect(C->sas, callback_connect_done, C)) == NULL)
+			goto err2;
+	} else {
+		if ((C->connect_cookie =
+		    network_connect_bind(C->sas, sa_b, callback_connect_done, C)) == NULL)
```
gperciva, line 251: Hmm.  If we changed the network_connect API, these 9 lines could simply be:

```
if ((C->connect_cookie = network_connect_full(C->sas, C->sa_b, NULL, callback_connect_done, C)) == NULL)
    goto err2;
```

@cperciva: the idea is to rename `network_connect_internal` to `network_connect_2` or `network_connect_full` or something like that.  (And expose the renamed `network_connect_internal` in the public API, of course.)

I'm concerned that if we add `network_connect_bind` now, then what happens if we ever want to add `timeo` to that?  Call it `network_connect_bind_timeo`, and then have 4 different functions for opening a network connection?

spipe/main.c
```diff
@@ -306,29 +326,33 @@ main(int argc, char * argv[])
 	if (ET.connection_error)
 		goto err0;
 
+	sock_addr_freelist(sas_b);
```
gperciva, line 329: Shouldn't this be in the "Clean up" section, on line 324?  (i.e. move it 5 lines higher)

lib/proto/proto_conn.c
nerijus: Fixed.

lib/proto/proto_conn.c
nerijus: Fixed.

spipe/main.c
nerijus: Fixed.

lib/proto/proto_conn.c
gperciva: Now that #354 has been merged, this can now be:

```
if ((C->connect_cookie = network_connect_bind(C->sas, C->sa_b, NULL, callback_connect_done, C)) == NULL)
    goto err2;
```

(#355 clarifies that this is safe, but the actual code isn't changing in that PR so it should be fine to change this line now.)

lib/proto/proto_conn.h
```diff
@@ -15,22 +15,23 @@ enum {
...
+ *     timeo, callback_dead, cookie):
  * Create a connection with one end at ${s} and the other end connecting to
- * the target addresses ${sas}.  If ${decr} is 0, encrypt the outgoing data;
- * if ${decr} is nonzero, decrypt the incoming data.  If ${nopfs} is non-zero,
- * don't use perfect forward secrecy.  If ${requirepfs} is non-zero, drop
- * the connection if the other end tries to disable perfect forward secrecy.
- * Enable transport layer keep-alives (if applicable) on both sockets if and
- * only if ${nokeepalive} is zero.  Drop the connection if the handshake or
+ * the target addresses ${sas}.  Bind outgoing address to ${sa_b} if it is
+ * not NULL.  If ${decr} is 0, encrypt the outgoing data;  if ${decr} is
```
gperciva, line 22: There's two spaces in "data;  if".  That should only be 1 space, please.

spiped/dispatch.c
```diff
@@ -199,25 +200,25 @@ callback_gotconn(void * cookie, int s)
...
+ * (if applicable) if and only if ${nokeepalive} is zero.  Drop connections if the
+ * handshake or connecting to the target takes more than ${timeo} seconds.  If
  * dispatch_request_shutdown() is called then ${conndone} is set to a non-zero
  * value as soon as there are no active connections.  Return a cookie which can
  * be passed to dispatch_shutdown() and dispatch_request_shutdown().
  */
 void *
 dispatch_accept(int s, const char * tgt, double rtime, struct sock_addr ** sas,
-    int decr, int nopfs, int requirepfs, int nokeepalive,
+    const struct sock_addr * sa_b, int decr, int nopfs, int requirepfs, int nokeepalive,
```
gperciva, line 221: Unfortunately this line ends up being 96 characters long; however, the `STYLE` file in the top-level directory calls for lines to "generally be 78 characters, and not more than 80 characters".

Please move `const struct sock_addr *` to the next line, and then the `size_t` onto the line after that.

Please also rewrap the function comment, as some of those lines are move than 80 chars.

lib/proto/proto_conn.h
nerijus: Fixed.

lib/proto/proto_conn.c
nerijus: Done.

nerijus:
Replying to https://github.com/Tarsnap/spiped/issues/351#issuecomment-1110190847
> Good catch about the port number. We could probably add `:0` to the bind address if it doesn't have a port number specified, to tell the OS to pick any port number it likes.

I should probably do this also; could someone please guide me where?

nerijus:
I removed `libcperciva/*` changes from PR, as they are merged already.

cperciva:
> Replying to [#351 (comment)](https://github.com/Tarsnap/spiped/issues/351#issuecomment-1110190847)
> 
> > Good catch about the port number. We could probably add `:0` to the bind address if it doesn't have a port number specified, to tell the OS to pick any port number it likes.
> 
> I should probably do this also; could someone please guide me where?

Sorry, I totally forgot about this.  I would say in the two `main` functions:

1. Look for the final `:` character in the bind address (using `strrchr`),
2. Check that everything after that is a digit.

If there is no `:` or there is a non-digit character following it (as in the case of `[ipv6::addr]`) then use `asprintf` to construct a string with the `:0` appended.

cperciva:
While I'm thinking about bind addresses: Do we fail with a sane error message if someone tries to bind to a unix socket address and connect over TCP, or bind to an IP address and connect to a unix socket?

gperciva:
> Sorry, I totally forgot about this. I would say in the two `main` functions:

What would you think about having a separate function in `main.c`, instead of the code being inside the `main()`?  I'm thinking of something like

```
static struct sock_addr **
required_sock_resolve(const char * addr, int add_zero_port)
```

that function would then be used for both `opt_b` and `opt_t`, and the same function.

... of course, that ~~begs~~ raises the question of whether it's worth adding such a function to `sock.h`, perhaps called `sock_resolve_required()`.

gperciva:
On the topic of a theoretical `sock_resolve_required`...

`perftests/recv-zeros/main.c` does:
```
if ((sas = sock_resolve(addr)) == NULL)
    goto err1;

if ((socket = sock_listener(sas[0])) == -1)
    goto err2;
```

However, `sock_listener()` doesn't check that its argument is not NULL; it jumps immediately to dereferencing `sa->ai_family`.

1) should `sock_listener()` check for a `NULL`, or is it a violation of the API to pass it a `NULL`?

2) I'm quite willing to believe that I screwed up when writing `perftests/recv-zeros/main.c` by not checking for `sas[0] == NULL`, but in which scenario would we want to have `sock_resolve()` return successfully with an empty list?

spiped/main.c
```diff
@@ -274,50 +284,92 @@ main(int argc, char * argv[])
 			warnp_syslog(1);
 	}
 
+	/* Resolve bind address. */
+	if (opt_b) {
+		char * a = strchr(opt_b, ':');
```
gperciva, line 289: Sorry for the delay, and thanks for this addition!  I've extracted the "add a :0 if necessary" into a new PR #358, since we'll need to use this code for both spiped/main.c and spipe/main.c.

spiped/main.c
```diff
@@ -274,50 +284,92 @@ main(int argc, char * argv[])
...
+		char * b = strrchr(opt_b, ':');
+		char * c = strrchr(opt_b, ']');
+		/* If there is no ':', it's ipv4 address (or Unix domain socket). */
+		if (b == NULL) {
+			if (asprintf(&bind_addr, "%s:0", opt_b) == -1) {
```
gperciva, line 294: Do we want to add :0 to a Unix domain socket?  I thought that we'd want to leave a Unix domain socket address unchanged.
(that's what I did in #358, but I might be wrong)

spiped/main.c
```diff
@@ -274,50 +284,92 @@ main(int argc, char * argv[])
...
+		/* If there are two different ':', it's ipv6 address. */
+		else if (a != b) {
+			/* If there is no ']', we need to add '[]' and ':0'. */
+			if (c == NULL) {
+				if (asprintf(&bind_addr, "[%s]:0", opt_b) == -1) {
+				    warnp("asprintf");
+				    goto err2;
+				}
+			}
+			/* If there is nothing after ']', we need to add ':0'. */
```
gperciva, line 308: If the ipv6 address already contains a port number, then `bind_addr` is unset.
(I've fixed this in #358)

gperciva:
Hi @nerijus,

I extracted the "add a port number if necessary" code into #358, and Colin added some review comments to it.  Could you please continue the work on `sock_addr_ensure_port()`?

nerijus:
OK, will do it a bit later.

spiped/main.c
```diff
@@ -274,50 +285,69 @@ main(int argc, char * argv[])
 			warnp_syslog(1);
 	}
 
+	/* Resolve bind address. */
+	if (opt_b) {
+		if ((bind_addr = sock_addr_ensure_port(opt_b))  == NULL) {
+			warnp("Failed to get bind address");
+			goto err2;
+		}
+		while ((sas_b = sock_resolve(bind_addr)) == NULL) {
```
gperciva, line 294: Thanks!  If you rebase onto master, you'll find a new `sock_resolve_one()`, which can simplify this code.

I'm also going to submit a further modification to `sock_resolve_one()`, which would include the `sock_addr_ensure_port()` functionality in it.  I don't know if that proposal will be accepted, though.  I'll get back to you when I have news.

spiped/main.c
```diff
@@ -274,50 +285,69 @@ main(int argc, char * argv[])
 			warnp_syslog(1);
 	}
 
+	/* Resolve bind address. */
```
gperciva, line 288: If this section occurred after resolving `sas_t`, then we'd have fewer modified lines, which always makes it easier to read & check commits.

gperciva: Come to think of it... if we have the modified `sock_resolve_one()`, then the `errX` labels would only increase by 1 instead of 2 for this code block.  In general, I think that's a good thing.

However, it might be annoying for you to modify all this code twice.  So maybe hold off on this change until the libcperciva infrastructure is finished.

gperciva:
Hi @nerijus, I'll take care of Colin's concerns about `sock_addr_ensure_port()`, and any other libcperciva changes that might be required.

spiped/main.c
nerijus: But there is no `sock_addr_ensure_port()` in  master...

gperciva: Thanks, and sorry for the delay.  There's now a `sock_resolve_one()` in master, which will take care of calling `sock_addr_ensure_port`.

gperciva:
Hi @nerijus, I'll take care of the rest of this PR, since the remaining stuff is all annoying nitpicky stuff.  Thanks for all your work on this!

gperciva:
Hi @nerijus, we've now merged this functionality in #364.  Thanks for all your work on this!

PR #354, opened 2022-04-26
Add network_connect_bind()

gperciva:


gperciva:
Hi @nerijus,

Would you mind splitting the libcperciva-specific changes from #353 into a separate commit?  That will make our testing easier.

This branch contains the libcperciva parts of #353 (all rebased together), plus three more commits that I added to fix problems.  If you don't mind splitting up your PR, please rebase all my commits into yours.

libcperciva/network/network_connect.c
```diff
@@ -15,7 +15,7 @@ struct connect_cookie {
 	int (* callback)(void *, int);
 	void * cookie;
 	struct sock_addr * const * sas;
-	struct sock_addr * sa_b;
+	const struct sock_addr * sa_b;
```
gperciva, line 18: This resolves a compiler warning, but I haven't thought about what it's actually doing.  I'm not at all certain that this is the correct fix.

gperciva: Here's the relevant error (without that commit):

```
../libcperciva/network/network_connect.c:227:10: error: assigning to 'struct sock_addr *' from 'const struct sock_addr *' discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
        C->sa_b = sa_b;
                ^ ~~~~
```

cperciva:
We can simplify the code by making `sock_connect_nb(sa)` just `sock_connect_bind_nb(sa, NULL)`.  (And then handling NULL in `sock_connect_bind_nb` by skipping the bind call.)

nerijus:
> Would you mind splitting the libcperciva-specific changes from #353 into a separate commit? That will make our testing easier.
> 
> This branch contains the libcperciva parts of #353 (all rebased together), plus three more commits that I added to fix problems. If you don't mind splitting up your PR, please rebase all my commits into yours.

I've rebased my PR on your branch `libcperciva-connect-bind`.

libcperciva/network/network_connect.c
```diff
@@ -118,8 +119,13 @@ tryconnect(struct connect_cookie * C)
 	/* Try addresses until we find one which doesn't fail immediately. */
 	for (; C->sas[0] != NULL; C->sas++) {
 		/* Can we try to connect to this address? */
-		if ((C->s = sock_connect_nb(C->sas[0])) != -1)
-			break;
+		if (C->sa_b != NULL) {
+			if ((C->s = sock_connect_bind_nb(C->sas[0], C->sa_b)) != -1)
+				break;
+		} else {
+			if ((C->s = sock_connect_nb(C->sas[0])) != -1)
```
gperciva, line 126: Hmm, we could make this whole section:
```
if ((C->s = sock_connect_nb(C->sas[0], C->sa_b)) != -1)
        break;
```
(and rely on `sock_connect_nb` handling C->sa_b being NULL or not)

That said, it could be argued that it would be clearer to see the `_bind_nb` vs. `_nb`.  I'm not convinced by that argument, but it's @cperciva's call.

libcperciva/util/sock.c
```diff
@@ -406,13 +406,35 @@ sock_connect(struct sock_addr * const * sas)
...
+{
+
+	/* Let sock_connect_bind_nb handle this. */
+	return (sock_connect_bind_nb(sa, NULL));
+}
+
+/**
+ * sock_connect_bind_nb(sa, sa_b):
+ * Create a socket, mark it as non-blocking, and attempt to connect to the
+ * address ${sa}.  If ${sa_b} is not NULL, then bind the socket to ${sa_b}.
```
gperciva, line 418: I think that this is what @cperciva asked for, but having a function called `sock_connect_bind_nb()` and then the comment `If ${sa_b} is not NULL, then bind the socket to ${sa_b}` strikes me as a bit weird.

If we're going with 3 different `network_connect` public functions (which all call `network_connect_internal`), then should we do a similar thing in `sock.c`?  i.e. add a `sock_connect_nb_internal`, which is called by `sock_connect_nb` and `sock_connect_bind_nb`?

libcperciva/network/network_connect.c
```diff
@@ -178,18 +182,26 @@ network_connect(struct sock_addr * const * sas,
...
-	return (network_connect_timeo(sas, NULL, callback, cookie));
+	/* Let network_connect_internal handle this. */
+	return (network_connect_internal(sas, NULL, NULL, callback, cookie));
 }
 
 /**
- * network_connect_timeo(sas, timeo, callback, cookie):
- * Behave as network_connect(), but wait a duration of at most ${timeo} for
- * each address which is being attempted.
+ * network_connect_internal(sas, sa_b, timeo, callback, cookie):
```
cperciva, line 190: If we move this function above `network_connect`, we won't need the forward declaration.  Slightly more churn in the file but it ends up being cleaner.

gperciva: Oh, I was aiming for minimal diffs.  Ok, I'll move it.

libcperciva/util/sock.c
cperciva: Let's keep it as-is for now.  If we end up creating a third "flavour" of `sock_connect_nb` then we can go down the `_internal` route at that point.

gperciva:
I believe that this is ready for final review and merging.

PR #355, opened 2022-05-09
Fix network_connect_bind() comment; sa_b can be NULL

gperciva:


gperciva:
Sorry, I just noticed this when re-reviewing #353.  The API comment for `network_connect_bind()` doesn't mention anything about what might happen if `sa_b` is NULL.

libcperciva/network/network.h
```diff
@@ -41,12 +41,13 @@ void * network_connect(struct sock_addr * const *,
...
 /**
  * network_connect_bind(sas, sa_b, callback, cookie):
- * Iterate through the addresses in ${sas}, attempting to create and connect
- * a non-blocking socket after binding to ${sa_b}.  Once connected, invoke
- * ${callback}(${cookie}, s) where s is the connected socket; upon fatal error
- * or if there are no addresses remaining to attempt, invoke
- * ${callback}(${cookie}, -1).  Return a cookie which can be passed to
- * network_connect_cancel() in order to cancel the connection attempt.
+ * Iterate through the addresses in ${sas}, attempting to create and connect a
+ * non-blocking socket.  If ${sa_b} is not NULL, then bind the socket to
```
cperciva, line 45: The binding happens before connecting, so this isn't right.

gperciva: Hmm.  It would be nice if the description of `network_connect_bind` remained fairly similar to that of `network_connect`.

How about "If sa_b is not NULL, then bind the socket to sa_b immediately after creating it"?

I've also modified the API comment for `sock_connect_bind_nb`, which has the same problem.

The alternative (if we don't add something like "immediately after creating it") for `sock_connect_bind_nb` would be:

> Create a socket.  If ${sa_b} is not NULL, then bind the socket to ${sa_b}.  Mark the socket as non-blocking, and attempt to connect to the address ${sa}.

I mean, sure that's a literal description of what's happening in the `.c` file, but it's rather awkward.

gperciva:
Rebased with an alternate wording.

cperciva:
Better, but what do you think of `s/then//` everywhere?  What you have is grammatical (if ... then ...) but ambiguous, since "then" could be interpreted as meaning "after completing the previous steps", and I think it's clear without that word.

gperciva:
Ok.  Revised the PR, and also clarified the commit message.

PR #356, opened 2022-05-13
Add sock_resolve_required

gperciva:


gperciva:
Draft of the `sock_resolve_required()` that I mentioned in #353.

I wasn't certain if I was going to like it, but having seen the changes to the various `main.c` files, I'm feeling more positive about it.  Don't know if you'll like it.

PR #357, opened 2022-05-21
sock_resolve_one

gperciva:


gperciva:
New version of #356 with reduced scope, aimed at the source/bind case, not for the target address.

The idea is that this `sock_resolve_one` can optionally be expanded to handle appending a ":0" to the address (if necessary).

libcperciva/util/sock.c
```diff
@@ -296,6 +296,65 @@ sock_resolve(const char * addr)
...
+		warn0("No addresses found for %s", addr);
+		goto err1;
+	}
+
+	/* If there's more than one address, give a warning. */
+	if (sas[1] != NULL)
+		warn0("Using the first of multiple addresses found for %s",
+		    addr);
+
+	/* Allocate a structure. */
```
cperciva, line 327: Err... we have `sock_addr_dup` if you want to do it that way.  But why not just free the bits we don't want?

```
sa = sas[0];
for (p = &sas[1]; *p != NULL; p++)
    sock_addr_free(*p);
free(sas);
```

spiped/main.c
```diff
@@ -275,17 +275,13 @@ main(int argc, char * argv[])
 	}
 
 	/* Resolve source address. */
-	while ((sas_s = sock_resolve(opt_s)) == NULL) {
+	while ((sa_s = sock_resolve_one(opt_s)) == NULL) {
```
cperciva, line 278: This changes the behaviour from "error out if there are no addresses" to "keep looping forever".

libcperciva/util/sock.c
gperciva: Huh, I never noticed `sock_util.h`.

I agree that avoiding the malloc is a cleaner solution.

spiped/main.c
gperciva: Sorry, I'm not following.  I completely believe that I screwed something up, and/or we need to pass an extra argument to `sock_resolve_one` to handle this case elegantly, but I'm not seeing it.

With git master,
```
$ ./spiped/spiped -d -s 257.0.0.1:8001 -t 127.0.0.1:8000 -k /dev/null -F
spiped: Error looking up 257.0.0.1: Name does not resolve
spiped: Error resolving socket address: 257.0.0.1:8001
$ ./spiped/spiped -d -s 257.0.0.1:8001 -t 127.0.0.1:8000 -k /dev/null -F -D
spiped: Error looking up 257.0.0.1: Name does not resolve
spiped: Error looking up 257.0.0.1: Name does not resolve
spiped: Error looking up 257.0.0.1: Name does not resolve
spiped: Error looking up 257.0.0.1: Name does not resolve
^C
```

(^C indicates me manually pressing ctrl-C, of course)

With this PR,
```
$ ./spiped/spiped -d -s 257.0.0.1:8001 -t 127.0.0.1:8000 -k /dev/null -F
spiped: Error looking up 257.0.0.1: Name does not resolve
spiped: Error resolving socket address: 257.0.0.1:8001
spiped: Error resolving socket address: 257.0.0.1:8001
$ ./spiped/spiped -d -s 257.0.0.1:8001 -t 127.0.0.1:8000 -k /dev/null -F -D
spiped: Error looking up 257.0.0.1: Name does not resolve
spiped: Error resolving socket address: 257.0.0.1:8001
spiped: Error looking up 257.0.0.1: Name does not resolve
spiped: Error resolving socket address: 257.0.0.1:8001
spiped: Error looking up 257.0.0.1: Name does not resolve
spiped: Error resolving socket address: 257.0.0.1:8001
spiped: Error looking up 257.0.0.1: Name does not resolve
spiped: Error resolving socket address: 257.0.0.1:8001
^C
```

Granted, the double error message in the first case isn't ideal.  But the "loop / no loop" behaviour looks the same?

cperciva: Hmm, I was thinking of the case where we successfully get zero addresses.  (i.e. `sock_resolve` returns `sas != NULL` but `sas[0] == NULL`.)

But I'm not even sure that's possible, so let's not worry about that case.

cperciva:
Change looks good, please rebase.

gperciva:
Ready for merge.

PR #358, opened 2022-05-31
Add sock_util_ensure_port()

gperciva:


gperciva:
I extracted this from #353, and made a few style fixes and minor code fixes.

I'm not certain about the name.  I think it makes sense to be in `sock_util.h`, since we'll need this for both spiped and spipe.  I opted for `sock_util_` instead of `sock_addr_` since this operates on strings rather than `struct sock_addr`.  However, I kind-of drew a blank for the rest of the name.  I ended up with `_ensure_port()`, but that was more because I had to name it something.

libcperciva/util/sock_util.c
```diff
@@ -276,3 +278,74 @@ sock_addr_prettyprint(const struct sock_addr * sa)
...
+
+	/* Get info about $addr. */
+	a = strchr(addr, ':');
+	b = strrchr(addr, ':');
+	c = strrchr(addr, ']');
+
+	/* Figure out what type of address $addr is. */
+	if (addr[0] == '/') {
+		/* If the address starts with '/', it's a Unix domain socket. */
+		return (strdup(addr));
```
cperciva, line 307: Good, but I would probably add "and doesn't need a port number" to the comment.

libcperciva/util/sock_util.c
```diff
@@ -276,3 +278,74 @@ sock_addr_prettyprint(const struct sock_addr * sa)
...
+	} else if (a == NULL) {
+		/* If there is no ':', it's an ipv4 addr that needs a ":0". */
```
cperciva, line 309: The string "localhost" is not an IPv4 address.

libcperciva/util/sock_util.c
```diff
@@ -276,3 +278,74 @@ sock_addr_prettyprint(const struct sock_addr * sa)
...
+		if (asprintf(&bind_addr, "%s:0", addr) == -1) {
+			warnp("asprintf");
+			goto err0;
+		}
+	} else if (a == b) {
+		/*
+		 * If there's a single ':', then it's an ipv4 addr which already
```
cperciva, line 316: Same goes for "www.tarsnap.com:8022".

libcperciva/util/sock_util.c
```diff
@@ -276,3 +278,74 @@ sock_addr_prettyprint(const struct sock_addr * sa)
...
+		 * contains a port.
+		 */
+		return (strdup(addr));
+	} else {
+		/* If there are two different ':', it's ipv6 address. */
+
+		/* If there is no ']', we need to add "[]" and ":0". */
```
cperciva, line 323: No no no.  The user MUST place IP addresses inside `[]`.  If they screw that up, they get what they get.  If we say "well we know what the user meant" here then we just train users to get it wrong in other places too.

libcperciva/util/sock_util.c
```diff
@@ -276,3 +278,74 @@ sock_addr_prettyprint(const struct sock_addr * sa)
...
+		if (c == NULL) {
+			if (asprintf(&bind_addr, "[%s]:0", addr) == -1) {
+				warnp("asprintf");
+				goto err0;
+			}
+		}
+		/* If there is nothing after ']', we need to add ":0". */
+		else if (c[1] == '\0') {
```
cperciva, line 331: Rather than looking for "stuff after the closing ]", check if `strspn(&b[1], "0123456789")` points at the terminating `\0` byte.  If it doesn't, add `:0`.

gperciva:
Ok, but:
- is this the right file?
- is this the right function name?

cperciva:
Yes it's the right file.  Anything uses asprintf (or relies on code in any other `.c` file aside from `warnp.c`) should be in `sock_util.c` rather than `sock.c`.

Function name... hmm.  Everything here is `sock_addr_` so maybe `sock_addr_add_port0`?  Or `sock_addr_ensure_port`, either works for me.

cperciva:
> > Everything here is sock_addr_ so maybe sock_addr_add_port0?
> 
> Yes, but everything else acts on `struct sock_addr *`; whereas this function only acts on strings.

Meh.  Sure, we're taking a string and returning a string, but it's a string which is going to be passed to other `sock_addr_*` functions.

PR #363, opened 2022-07-30
Proto bind

gperciva:


gperciva:
This is split from #353; let's get all the library code sorted out and nitpick the comments.

PR #369, opened 2022-08-28
Libcperciva import

gperciva:


gperciva:
This includes the fix to use clang's `-Wreserved-id-macro` flag https://github.com/Tarsnap/libcperciva/pull/451.

PR #373, opened 2022-12-05
Add wait_ms()

gperciva:


gperciva:
This PR is in the gray zone of "it's only test code, just copy it again" and "let's shove it into a library".  We currently have 3 `static int wait_ms()` functions in spiped; this consolidates them.

I'm mostly interested for ease of debugging: when I wonder "what would happen if X took longer to return?", it's annoying to copy that function into a new file.  I'd rather just shove an `#include "wait_ms.h"` and then `wait_ms(100)` somewhere.

Also, if we wanted to make `wait_ms` fancier -- such as making it robust against `EINTR` -- that would be easier if we only had one copy of it floating around.

cperciva:
I would (a) call this millisleep (in analogy with nanosleep), and (b) declare it `static inline` in the `.h` file.  It's not worth having a separate `.c` file for one trivial function like this.

gperciva:
Rebased accordingly.

PR #384, opened 2023-06-06
libcperciva import

gperciva:

PR #385, opened 2023-06-06
Tests faster

gperciva:


gperciva:
Includes #384, in case you wanted to see how `wait_while()` was going to be used.

Drops the time to run the spiped tests from 27s to 22s.  A notable chunk of that time (10s) is spent on `08-send-data-spipe-limited-bps`, which (deliberately) only operates at 1000 bytes per second.

Amusingly, most tests end so quickly that valgrind needs extra hard-coded time to finish writing the log files for the servers.  I'm working on a better fix for that.

PR #393, opened 2023-10-02
Libcperciva import and trivial fixes

gperciva:


gperciva:
Includes function pointer style in `lib/dnsthread/dnsthread.c`.

Issue #397, opened 2024-01-31
Segfault: Error running event loop, Bad file descriptor


trmz:
Compiled latest (22cc9b4c18) on a couple different versions of Debian. 
Segfaults pretty consistently, after the first one or two or few tunneled connections (to PostgreSQL, for the curious).
(Here, I'm aware -b 127.0.0.1 is unnecessary, but it segfaults all the same regardless of the chosen address.)

Command:
```spiped -d -k /root/test.key -p /root/test.pid -s [192.168.1.2]:7700 -t [127.0.0.1]:5432 -b 127.0.0.1 -n 40```

Error:
```
spiped: shutdown: Transport endpoint is not connected
spiped: Error running event loop
spiped: close: Bad file descriptor
```

Syslog:
```
Jan 30 21:04:21 host kernel: [11486873.888877] spiped[7300]: segfault at 29 ip 0000000000407805 sp 00007ffeefb62fb0 error 4 in spiped-22cc9b4c18[400000+17000]
Jan 30 21:04:21 host kernel: [11486873.889835] audit: type=1701 audit(1706666661.707:67894): auid=1011 uid=0 gid=0 ses=27243 pid=7300 comm="spiped" exe="/usr/local/bin/spiped-22cc9b4c18" sig=11
```

Here's a segfault from a different server, Debian 10 :
```
[  +0.036077] spiped[12592]: segfault at a9 ip 0000559136791e2c sp 00007fff2bfef820 error 4 in spiped-22cc9b4c18[55913678c000+11000]
[  +0.009377] Code: 58 fe ff ff bb ff ff ff ff e9 4e fe ff ff 48 83 ec 08 be 00 00 00 00 e8 71 fd ff ff 48 83 c4 08 c3 48 85 ff 74 17 53 48 89 fb <48> 8b 7f 08 e8 7b a4 ff ff 48 89 df e8 73 a4 ff ff 5b c3 c3 41 57
```

EDIT: 
Nevermind, I lied.
Thought it worked fine without option -b ... turns out it was just luck, segfaults all the same with or without -b.  :/
Feel free to rename this issue, or I can create a new one. 
Note that spiped v1.6.2 has been working stably (but lacks the -b option I need).

cperciva:
Can you get a backtrace from the segfault?

trmz:
Hey, I'd like to, but haven't ever tried that. 
Would following this guide produce usable output ? 
https://serverfault.com/a/61676

[...]

PR #398, opened 2024-02-03
Fixes

gperciva:


gperciva:
Two embarrassing mistakes here; the one in `dispatch_shutdown()` is particularly bad.

These were uncovered during the investigation into #397, but they don't resolve whatever is happening in that issue.  (I can reproduce the problem, but I'm still trying to figure out exactly what's happening there.)

gperciva:
Added a third fix (not as embarrassing).

References (but nothing looks relevant):
- add LIBS to the build system was originally https://github.com/Tarsnap/libcperciva/pull/474
- using LIST for `dispatch.c` https://github.com/Tarsnap/spiped/pull/53

More importantly, we discussed `shutdown()` receiving an `ENOTCONN` back in https://github.com/Tarsnap/spiped/pull/350; the final opinion was:
> Ok, I *think* this is ok; it looks to me like ENOTCONN should not (non-erroneously) happen here, even if we get RST.
>
> If users start reporting ENOTCONN then we should probably treat it as a non-error.

We now have a report of ENOTCONN in #397 (although the cause was disguised by the problem in `dispatch_shutdown()`), so this PR makes it a non-error.

(I can reproduce the ENOTCONN using postgresql; or with our test suite if I add a few `sleep(1)` to our code.  I haven't yet been able to reproduce ENOTCONN with test programs without modifying spiped code, but I think it's sufficiently clear what's happening.)

gperciva:
Oh, in case it's relevant for ENOTCONN: original report on Linux, but I reproduced it on FreeBSD 14.0-RELEASE-p3.

PR #401, opened 2024-03-10
dispatch_shutdown(): clarify conn_cookies clean-up

gperciva:


gperciva:
I'm not wild about adding code just to address false positives from static analyzer(s), but we have the same construct in `tests/nc-server/simple_server.c`, so there *is* some precedent.

PR #407, opened 2024-08-03
Invoke usage for missing hostnames in -[st]

cperciva:
A source or destination address of the form ":1234" is parsed as a hostname of "" which is guaranteed to never work.  While we can't reject all possible invalid addresses at this point -- and probably shouldn't since the resolver can give more informative errors -- this particular case can easily be produced from a broken shell script (with a hostname variable accidentally not set), so it's more worthy of rejection via usage() than most.

Suggested by:	Ross Richardson

### DIFF
--- a/spipe/main.c
+++ b/spipe/main.c
@@ -221,7 +221,7 @@ main(int argc, char * argv[])
 		usage();
 	if (!(opt_o > 0.0))
 		usage();
-	if (opt_t == NULL)
+	if ((opt_t == NULL) || (opt_t[0] == ':'))
 		usage();
 
 	/* Initialize the "events & threads" cookie. */
--- a/spiped/main.c
+++ b/spiped/main.c
@@ -239,9 +239,9 @@ main(int argc, char * argv[])
 		usage();
 	if ((opt_r != 60.0) && opt_R)
 		usage();
-	if (opt_s == NULL)
+	if ((opt_s == NULL) || (opt_s[0] == ':'))
 		usage();
-	if (opt_t == NULL)
+	if ((opt_t == NULL) || (opt_t[0] == ':'))
 		usage();
 
 	/*


cperciva:
@gperciva Quick sanity-check to make sure I didn't typo anything please.

spipe/main.c
```diff
@@ -221,7 +221,7 @@ main(int argc, char * argv[])
 		usage();
 	if (!(opt_o > 0.0))
 		usage();
-	if (opt_t == NULL)
+	if ((opt_t == NULL) || (opt_t[0] == ':'))
```
gperciva, line 224: If a shell variable isn't set, how would this help me track down the problem?  I'd say that
```
$ ./spipe/spipe -t ":1234" -k /dev/null
spipe: Error looking up : Name does not resolve
spipe: Error resolving socket address: :1234
```
is a somewhat more helpful message than
```
$ ./spipe/spipe -t ":1234" -k /dev/null
usage: spipe -t <target socket> -k <key file> [-b <bind address>] [-f | -g]
    [-j] [-o <connection timeout>]
       spipe -v
```

Now, if we wanted to add a function that checked for this and gave a more specific error message, I'd be all for it.  (Probably as a `static int`, rather than a separate header & c file.)

cperciva: Well, the main reason is for `spiped` because there the process could end up running forever with DNS lookups failing.  I agree that it's less useful for `spipe` since it will already fail promptly.

gperciva: My point is that `usage()` doesn't help to diagnose the problem, whereas the status quo does (somewhat).

If you want to give a better hint, how about:
```
if (addr[0] == ':') {
        warn0("No hostname in \"%s\"", addr);
        goto err0;
}
```

to `sock_resolve_host()` line 87 (just above the `/* Perform DNS lookup. */`)?  That would give a more informative error message, and also benefit all libcperciva software.

gperciva: Oops, that error check doesn't work (since the `addr:host` separation has already taken place).

`sock_resolve()` is better than `sock_resolve_host()`.  I uploaded that as #408 so you can see it more easily.

cperciva:
Closing since this was addressed in #408 .

PR #408, opened 2024-08-04
Give an error for addresses lacking a hostname

gperciva:
A source or destination address of the form ":1234" is parsed as a hostname of "" which is guaranteed to never work.  While we don't want to give special error messages for all possible invalid addresses, this particular case can easily be produced from a broken shell script (with a hostname variable accidentally not set), so it's worth having a special error message.

Suggested by:	Ross Richardson

### DIFF
--- a/libcperciva/util/sock_util.c
+++ b/libcperciva/util/sock_util.c
@@ -11,6 +11,7 @@
 
 #include "asprintf.h"
 #include "sock.h"
+#include "warnp.h"
 
 #include "sock_internal.h"
 #include "sock_util.h"
@@ -334,4 +335,36 @@ sock_addr_ensure_port(const char * addr)
 err0:
 	/* Failure! */
 	return (NULL);
+}
+
+/**
+ * sock_addr_validate(addr):
+ * Check that ${addr} is syntactically valid, but do not perform any address
+ * resolution.
+ */
+int
+sock_addr_validate(const char * addr)
+{
+
+	/* Sanity check. */
+	assert(addr != NULL);
+
+	/* Check for an empty address. */
+	if (strlen(addr) == 0) {
+		warn0("Empty socket address.");
+		goto err0;
+	}
+
+	/* If this isn't a UNIX socket address, check for a missing hostname. */
+	if ((addr[0] != '/') && (addr[0] == ':')) {
+		warn0("No host in \"%s\"", addr);
+		goto err0;
+	}
+
+	/* Success! */
+	return (0);
+
+err0:
+	/* Failure! */
+	return (-1);
 }
--- a/spipe/main.c
+++ b/spipe/main.c
@@ -13,6 +13,7 @@
 #include "graceful_shutdown.h"
 #include "parsenum.h"
 #include "sock.h"
+#include "sock_util.h"
 #include "warnp.h"
 
 #include "proto_conn.h"
@@ -221,7 +222,9 @@ main(int argc, char * argv[])
 		usage();
 	if (!(opt_o > 0.0))
 		usage();
-	if (opt_t == NULL)
+	if ((opt_t == NULL) || sock_addr_validate(opt_t))
+		usage();
+	if (opt_b && sock_addr_validate(opt_b))
 		usage();
 
 	/* Initialize the "events & threads" cookie. */
--- a/spiped/main.c
+++ b/spiped/main.c
@@ -13,6 +13,7 @@
 #include "parsenum.h"
 #include "setuidgid.h"
 #include "sock.h"
+#include "sock_util.h"
 #include "warnp.h"
 
 #include "dispatch.h"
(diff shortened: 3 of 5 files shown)


gperciva:
This is an alternate solution to #407.  (The commit message was tweaked accordingly.)

gperciva:
Example of message:
```
$ ./spipe/spipe -t ":1234" -k /dev/null
spipe: No hostname in ":1234"
spipe: Error resolving socket address: :1234
```

cperciva:
Ok, this adds a more helpful error message, but doesn't solve the problem that `spiped -e -s '[127.0.0.1]:12345' -t ':12346' -k /COPYRIGHT -D` daemonizes before it performs a DNS lookup and then spews error messages forever.

Can you write a `sock_validate` function which checks if an address is syntactically valid (i.e. if it *might* work) without actually doing the address resolution?  Then we can call that from `spiped`.

gperciva:
Oh!  Oops, ok, I get it now.  Will do.

gperciva:
Updated.

I also added a check for `opt_b`.

libcperciva/util/sock_util.c
```diff
@@ -335,3 +336,36 @@ sock_addr_ensure_port(const char * addr)
...
+ * Check that ${addr} is syntactically valid, but do not perform any address
+ * resolution.
+ */
+int
+sock_addr_validate(const char * addr)
+{
+
+	/* Sanity checks. */
+	if (addr == NULL) {
+		warn0("Empty sock address.");
```
cperciva, line 351: s/sock/socket/; we use `sock` in the code for conciseness, but `socket` and `socket address` should be used in user-facing places.

libcperciva/util/sock_util.c
```diff
@@ -335,3 +336,36 @@ sock_addr_ensure_port(const char * addr)
...
+		goto err0;
+	}
+	if (strlen(addr) == 0) {
+		warn0("Empty sock address.");
+		goto err0;
+	}
+
+	/* Check for a missing hostname. */
```
cperciva, line 359: UNIX socket addresses don't have hostnames.  The code is actually fine since UNIX socket addresses will start with `/` so anything with `addr[0] == ':'` isn't a valid UNIX *or* network socket address, but the comment (and possibly the error message) needs to be clarified.

libcperciva/util/sock_util.c
```diff
@@ -335,3 +336,36 @@ sock_addr_ensure_port(const char * addr)
...
+ * sock_addr_validate(addr):
+ * Check that ${addr} is syntactically valid, but do not perform any address
+ * resolution.
+ */
+int
+sock_addr_validate(const char * addr)
+{
+
+	/* Sanity checks. */
+	if (addr == NULL) {
```
cperciva, line 350: On second thought, I'm not a big fan of calling `NULL` an "empty" address.  It's not an address at all.  Also, in combination with the spiped code, forgetting to specify the address results in "Empty sock(et) address." followed by usage().

I would omit the `NULL` check from here and have the user responsible for checking for NULLity.

gperciva: ... how about an `assert(addr != NULL)`?

I mean, this is shared code for a function called `sock_addr_validate()`.  Moving a sanity check out to user code seems an odd thing to do, especially since the function is only a handful of lines.

Yes, given the current design of spiped we should never get a `NULL`, but what about future use in other projects?

If you're sure you don't want even an `assert`, I'll nuke it quickly.

gperciva:
Revised as requested, with the exception of an `assert()`.

cperciva:
If you can make that last change and clean up the commit history, I'll rebase this (and then you can cherry pick the libarchive commit over to that repo).

gperciva:
Rebased, ready for merge.

spipe/main.c
```diff
@@ -221,7 +222,9 @@ main(int argc, char * argv[])
 		usage();
 	if (!(opt_o > 0.0))
 		usage();
-	if (opt_t == NULL)
+	if (sock_addr_validate(opt_t))
```
cperciva, line 225: `if (opt_t == NULL || sock_addr_validate(opt_t))`

spiped/main.c
```diff
@@ -239,9 +240,11 @@ main(int argc, char * argv[])
 		usage();
 	if ((opt_r != 60.0) && opt_R)
 		usage();
-	if (opt_s == NULL)
+	if (sock_addr_validate(opt_s))
```
cperciva, line 243: As above.

gperciva:
Fixed.  (also `opt_t`, just below `opt_s` in `spiped/main.c`)

cperciva:
Thanks, please clean up the history.

gperciva:
Rebased, ready for  merge.

PR #413, opened 2025-01-13
Generalize ipc_sync to allow threads as well; use with standalone-enc

gperciva:


gperciva:
(I'm not certain if this is a neat idea, or over-engineered madness.)

I wanted a semaphore abstraction (wait & signal) that worked with processes and threads.  (POSIX semaphores aren't supported on macOS.)

We can't use ipc_sync directly in threads, because ipc_sync closes one read and write end of the pipe.  So this PR omits those parts in the "threaded version" of `ipc_sync.c`.

(If you think this is an interesting idea but now's not the time, I'll close this PR and submit it to libcperciva in a week or two.)

gperciva:
On second thought, this _should_ go to libcperciva first, with a test.

PR #414, opened 2025-01-14
standalone-enc: move fd_drain() to a separate file

gperciva:


gperciva:
It's useful to share `fd_drain()` -- or rather, `fd_drain_fork()` -- between multiple tests.

At the moment it's only used in `_transfer_noencrypt.c` and `_pipe_socketpair_one.c`, but we'll be adding `_pipe_socketpair_two.c` soon.

perftests/standalone-enc/fd_drain.h
```diff
@@ -0,0 +1,24 @@
...
+
+/* The smallest this can be is PCRYPT_ESZ (which is 1060). */
+#define FD_DRAIN_MAX_SIZE 16384
+
+/**
+ * fd_drain(fd_p):
+ * Drain bytes from the file descriptor ${*fd_p} -- passed as an (int *) -- as
+ * quickly as possible.
+ */
+int fd_drain(void *);
```
gperciva, line 14: My original intent was only to expose `fd_drain_fork()`, but `standalone_transfer_noencrypt.c` uses threads rather than processes, so it needs to call `fd_drain()` directly.

gperciva:
This has no impact on the perftest timings.

When I first glanced at it with N=5 it looked like it did, but repeating the test with N=11 and then N=31 showed it was a fluke.

perftests/standalone-enc/fd_drain.c
```diff
@@ -0,0 +1,63 @@
...
+fd_drain(void * cookie)
+{
+	int fd = *((int *)cookie);
+	uint8_t mybuf[FD_DRAIN_MAX_SIZE];
+	ssize_t readlen;
+
+	/* Loop until we hit EOF. */
+	do {
+		/*
+		 * This will almost always read 1060 bytes (size of an
```
cperciva, line 24: This comment seems like it made sense in the original context of where this code was used, but doesn't belong any more once `fd_drain` is in a separate file since it could end up being used by completely different callers.

perftests/standalone-enc/fd_drain.h
cperciva: I'm not a fan of the "pass a pointer to the file descriptor, cast to void *" API here.  How about `int fd_drain(int)` and `static int fd_drain_internal(void *)` which gets called by both of the externally-visible functions?

perftests/standalone-enc/fd_drain.h
```diff
@@ -0,0 +1,23 @@
+#ifndef FD_DRAIN_H_
+#define FD_DRAIN_H_
+
+#include <sys/types.h>
+
+/* The smallest this can be is PCRYPT_ESZ (which is 1060). */
```
cperciva, line 6: Not sure if this comment is necessary or even correct any more?

cperciva:
Flagged one more comment.  Aside from that, looks good; please tidy up the history.

perftests/standalone-enc/fd_drain.h
gperciva: Hmm, I guess that if it's something tiny like 10 bytes, that would just mean `read` gets called more often.

... yeah, I'll delete it.

gperciva:
Rebased, ready for merge.

PR #418, opened 2025-03-08
sock_addr_validate: remove redundant expression

gperciva:


gperciva:
Whoops, this slipped through the cracks.

I normally run `clang-tidy` for new features, but to save time I omit `libcperciva/` files from the check because I figure those files will be checked when I run `clang-tidy` in that repository.

Then when I was importing this change to `libcperciva`, I didn't bother with `clang-tidy` because it was only importing a commit, not "new" code.

cperciva:
The original code was redundant but was easier to understand -- "check for a missing hostname" doesn't make sense by itself since local socket paths don't have hostnames.  We can change the code if you want to keep clang-tidy happy but the comment should explain that local socket paths will pass the test since they start with `/`.

gperciva:
Ok, revised comment.  It feels a bit wordy, but I couldn't think of a shorter way without ambiguity.

In case you were wondering: I disable approximately 70 clang-tidy checks; the remainder are ones which I think could give useful warnings in some cases.

PR #419, opened 2025-03-24
Shutdown err handling

cperciva:


gperciva:
Added a REBASE commit which added some comments.

gperciva (approved):

cperciva:
I changed the comments, but thanks for pointing out that I needed to add some.

PR #427, opened 2026-07-12
proto_handshake: fix clean-up on mkkeys() failure

gperciva:


lib/proto/proto_handshake.c
```diff
@@ -304,9 +304,15 @@ handshakedone(struct handshake_cookie * H)
 	/* Return status code from callback. */
 	return (rc);
 
-err0:
+err1:
+	/* Clean up. */
+	if ((rc = handshakefail(H)) == 0) {
+		/* If there's no error in the clean-up, set a default error. */
+		rc = -1;
```
gperciva: I'm less confident about this than I am most of the other recent PRs -- why don't we want `handshakefail()` to return -1 in more cases?

cperciva: Yeah this bit isn't right.  An error from `proto_crypt_mkkeys` means "we were unable to generate keys because we ran out of memory"; it's a fatal error.  We call `handshakefail` in many non-fatal code paths, e.g. if the network dropped before we got DH parameters, so in the common case it will return 0, and your patch will mask the "we should exit rather than just dropping this one connection" fatal error from `proto_crypt_mkkeys`.

This should just be

```
err1:
    handshakefail(H);

    /* Failure! */
    return (-1);
```

lib/proto/proto_handshake.c
```diff
@@ -304,7 +304,10 @@ handshakedone(struct handshake_cookie * H)
 	/* Return status code from callback. */
 	return (rc);
 
-err0:
+err1:
+	/* Clean up, but we don't care if the callback succeeded or failed. */
```
gperciva: Is it worth adding a comment about this?  Or mention something about informing the caller?

cperciva: No, we have a lot of "we're cleaning up from an error, there's no point checking if we had another error" paths.  This might be an idiom which is worth documenting in STYLE I guess ("don't check for for failures in an error-handling path if they will not affect behaviour")?  But there's definitely no point documenting it every time it happens.

gperciva:
Rebased

PR #428, opened 2026-07-13
spiped: fix 'goto err' labels

gperciva:


gperciva:
Looking through the history, the first sketch of this commit in #353 required all the `errX` numbers to increment by 2.  When we added `sock_resolve_one()`, that meant the commit only required main.c's `errX` numbers to increment by 1, but evidently we missed two instances of that change.

The bug reporter notes that dereferencing a NULL pointer is a crash.  I'm guessing that you would classify this as a "harmless" bug.

PR #433, opened 2026-08-05
Fix network usage

gperciva:


gperciva:
This follows up to https://github.com/Tarsnap/libcperciva/pull/543.

tests/nc-server/simple_server.c
```diff
@@ -330,13 +333,10 @@ simple_server(const char * addr, size_t nconn_max, size_t shutdown_after,
 	/* Success! */
 	return (0);
 
-err4:
-	if (A->accept_cookie != NULL)
-		network_accept_cancel(A->accept_cookie);
 err3:
-	free(A);
+	simple_server_shutdown(A);
```
cperciva, line 337: Is this safe to call on all possible event-loop exits?

gperciva: Haven't seen a problem yet, but I'll return to checking it tomorrow.

gperciva:
Revised with 2 new commits, although I'd like to change the order before the final merge.  In particular, move the `do not re-launch any accept in shutdown` earlier so it's the first `nc-server` update.  (The `don't quit server` will stay at the end.)

tests/nc-server/simple_server.c
```diff
@@ -240,8 +240,13 @@ simple_server_shutdown(void * cookie)
 	struct conn_list_node * node_ptr;
 
 	/* Cancel any further accepts. */
-	if (A->accept_cookie != NULL)
+	if (A->accept_cookie != NULL) {
 		network_accept_cancel(A->accept_cookie);
+		A->accept_cookie = NULL;
+	}
+
+	/* Stop drop() -> conndied() -> doaccept() from arming a new accept. */
```
cperciva: I'd put this above the `/* Cancel any further attempts. */` block; we're single-threaded so there isn't actually any race, but it's cleaner in principal to say "don't restart this, and stop it now" rather than "stop it now and don't restart it".

cperciva:
s/it's is part/it's part/.

cperciva:
What's the "NRC" in "network_read_cancel(NRC)"?

lib/proto/proto_pipe.c
```diff
@@ -193,7 +193,13 @@ callback_pipe_read(void * cookie, int status)
 	/* Inform the upstream that our status has changed. */
 	return ((P->callback)(P->cookie));
 
-err0:
+err1:
+	/* Record that this connection is broken. */
+	*(P->status) = -1;
+
+	/* Inform the upstream that our status has changed. */
+	P->callback(P->cookie);
```
cperciva: `(P->callback)(P->cookie);` is our normal style for this.

cperciva:
A few minor nits, but I think the code is right in concept now.  Yes, please re-order commits to avoid introducing a bug and then fixing it.

gperciva:
> What's the "NRC" in "network_read_cancel(NRC)"?

network_read_cookie.

I'll reword it.

gperciva:
Next edition:
- two REBASE commits to fix issues you mentioned.  Also, I added `(void)` to the callbacks in err1 paths.
- in the process of checking them, Claude suggested a readability improvement to `simple_server_shutdown()` which I liked.

I haven't re-ordered the commits yet, because that would spoil the attempt to only show changes since your previous review.

cperciva:
Please tidy up the git history; the new commits look good but I'll take another look once you've rebased.

gperciva:
Rebased, ready for final review.

PR #437, opened 2026-09-06
proto_crypt: clear the remaining stack buffers

### LINKED ISSUE #436: [bug bounty] Key file contents and dhmac keys are left on the stack in proto_crypt.c

### Summary

`proto_crypt_secret()` leaves the entire contents of the key file in a stack
buffer, and `proto_crypt_dhmac()` leaves the derived diffie-hellman MAC keys in
another.  Neither is wiped on any path.

This is the same class as #426 / Tarsnap/tarsnap#709, which wiped `dk_2` and
`nonce_y` in `proto_crypt_mkkeys()`.  That fix covered `mkkeys()` only; these
two are the remaining unwiped buffers in the same file.

### 1. The key file contents stay on the stack

`lib/proto/proto_crypt.c`:

```c
struct proto_secret *
proto_crypt_secret(const char * filename)
{
	SHA256_CTX ctx;
	FILE * f;
	struct proto_secret * K;
	uint8_t buf[BUFSIZ];
	size_t lenread;
...
	/* Read the file until we hit EOF. */
	while ((lenread = fread(buf, 1, BUFSIZ, f)) > 0)
		SHA256_Update(&ctx, buf, lenread);
...
	/* Compute the final hash and wipe context state. */
	SHA256_Final(K->K, &ctx);

	/* Success! */
	return (K);
```

`buf` holds the last `BUFSIZ` bytes of the key file — the pre-image of the
shared secret, and for the common case of a key file of 8192 bytes or less,
the whole of it.  It is never zeroed, on the success path or on the
read-error path at `err2`.

The function already takes care over this material two lines above: the
`SHA256_Final()` call on the error path exists only to wipe the context, and
its comment says so.  `proto_crypt_secret_free()` likewise zeroes the derived
secret before freeing it.  The buffer the secret was derived *from* is the one
thing in the path that is left behind.

Both `spiped` and `spipe` call this once at startup and then run for the
lifetime of the process, so the frame is not reliably overwritten afterwards.

### 2. The diffie-hellman MAC keys stay on the stack

```c
void
proto_crypt_dhmac(const struct proto_secret * K, ...)
{
	uint8_t nonce_CS[PCRYPT_NONCE_LEN * 2];
	uint8_t dk_1[PCRYPT_DHMAC_LEN * 2];
...
	/* Compute dk_1. */
	PBKDF2_SHA256(K->K, 32, nonce_CS, PCRYPT_NONCE_LEN * 2, 1,
	    dk_1, PCRYPT_DHMAC_LEN * 2);

	/* Copy out diffie-hellman parameter MAC keys (in the right order). */
	memcpy(dhmac_c, &dk_1[0], PCRYPT_DHMAC_LEN);
	memcpy(dhmac_s, &dk_1[PCRYPT_DHMAC_LEN], PCRYPT_DHMAC_LEN);
}
```

`dk_1` is key material derived from `K->K`, and is left in the frame when the
function returns.  This is called once per connection.

### Fix

PR to follow: `insecure_memzero()` for `buf` on both exits from
`proto_crypt_secret()`, and for `dk_1` at the end of `proto_crypt_dhmac()`,
using the same idiom as #426.

Two buffers in this file are deliberately left alone:

* `ctx` in `proto_crypt_enc()` and `proto_crypt_dec()` — it is a copy of
  `k->ctx_init`, which lives in the key structure for the whole connection
  anyway, and those two functions run once per 1024-byte packet.
* `nonce_CS` in `proto_crypt_dhmac()` — the nonces are sent over the wire in
  the clear.

### How this was found

Source review of `lib/proto/proto_crypt.c` at current master, checking each
stack buffer that receives key material against the wipes added by #426.

woahwhattheheck:

Fixes #436.

#426 wiped `dk_2` and `nonce_y` in `proto_crypt_mkkeys()`.  Two buffers holding
key material in the same file were not covered by it:

* `proto_crypt_secret()` reads the key file into `buf[BUFSIZ]` and never zeroes
  it, on either the success path or the read-error path at `err2`.  That buffer
  is the pre-image of the shared secret — the whole key file, for a key file of
  `BUFSIZ` bytes or less.  The function already wipes the SHA-256 context two
  lines above, and `proto_crypt_secret_free()` zeroes the derived secret; the
  buffer it was derived from was the gap.  Both `spiped` and `spipe` call this
  once at startup and then run for the lifetime of the process.
* `proto_crypt_dhmac()` leaves `dk_1`, the derived diffie-hellman MAC keys, in
  the frame when it returns.  Called once per connection.

Three `insecure_memzero()` calls, using the idiom from #426.

### Deliberately not covered

* `ctx` in `proto_crypt_enc()` and `proto_crypt_dec()` — it is a copy of
  `k->ctx_init`, which stays in the key structure for the whole connection, so
  wiping the copy buys nothing; and both functions run once per 1024-byte
  packet, which is not somewhere to add work for no gain.
* `nonce_CS` in `proto_crypt_dhmac()` — the nonces are sent over the wire in
  the clear.

Happy to add either if you would rather have the file uniformly wiped.

### Testing

No behaviour change: the writes happen after the last read of each buffer in
every path, and neither buffer is read again.  `insecure_memzero.h` is already
included by this file.

### DIFF
--- a/lib/proto/proto_crypt.c
+++ b/lib/proto/proto_crypt.c
@@ -113,6 +113,9 @@ proto_crypt_secret(const char * filename)
 	/* Compute the final hash and wipe context state. */
 	SHA256_Final(K->K, &ctx);
 
+	/* Clear sensitive material from the stack. */
+	insecure_memzero(buf, sizeof(buf));
+
 	/* Success! */
 	return (K);
 
@@ -123,6 +126,9 @@ err2:
 
 	/* Wipe context state. */
 	SHA256_Final(K->K, &ctx);
+
+	/* Clear sensitive material from the stack. */
+	insecure_memzero(buf, sizeof(buf));
 err1:
 	proto_crypt_secret_free(K);
 err0:
@@ -166,6 +172,9 @@ proto_crypt_dhmac(const struct proto_secret * K,
 	/* Copy out diffie-hellman parameter MAC keys (in the right order). */
 	memcpy(dhmac_c, &dk_1[0], PCRYPT_DHMAC_LEN);
 	memcpy(dhmac_s, &dk_1[PCRYPT_DHMAC_LEN], PCRYPT_DHMAC_LEN);
+
+	/* Clear sensitive material from the stack. */
+	insecure_memzero(dk_1, sizeof(dk_1));
 }
 
 /**


tokenjunkielabs:
Following up on first-reporter bug bounty #436. This PR is the `woahwhattheheck` carrier for the unwiped key-file and DHMAC stack buffers. Please review the report/fix when convenient and classify #436 under the repository's published bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #439, opened 2026-09-06
dnsthread: register the wakeup listener before starting work

### LINKED ISSUE #438: [bug bounty] dnsthread_resolveone() abandons a running resolution if the wakeup listener fails to register

### Summary

If `events_network_register()` fails in `dnsthread_resolveone()`, the worker
thread has already been given the work and woken up.  The function returns -1,
but the resolution goes ahead with nothing registered to read its completion
byte.  That leaks the address and the resolved `sock_addr` list, never calls
the caller's callback, and leaves a byte in the wakeup socketpair which makes
the *next* resolution free a buffer while the worker thread is using it.

Same class as #427 / Tarsnap/tarsnap#710, in a different file.

### The ordering

`lib/dnsthread/dnsthread.c`:

```c
	/* Duplicate the address to be resolved. */
	if ((T->addr = strdup(addr)) == NULL)
		goto err1;

	/* Remember what callback we'll need to do eventually. */
	T->callback = callback;
	T->cookie = cookie;

	/* There is now work for the thread to do. */
	T->state = THREAD_HASWORK;

	/* Wake up the worker thread. */
	if ((rc = pthread_cond_signal(&T->cv)) != 0) {
		warn0("pthread_cond_signal: %s", strerror(rc));
		goto err1;
	}

	/* We want a callback when the worker thread pokes us. */
	if (events_network_register(callback_resolveone, T, T->wakeupsock[1],
	    EVENTS_NETWORK_OP_READ)) {
		warnp("Error registering wakeup listener");
		goto err1;
	}
...
err1:
	pthread_mutex_unlock(&T->mtx);
err0:
	/* Failure! */
	return (-1);
```

`err1` only unlocks the mutex.  By the time the registration is attempted the
state is already `THREAD_HASWORK` and the worker has been signalled, and that
cannot be taken back.

### What follows from one failed registration

**1. Two leaks.**  `callback_resolveone()` is the only thing that frees
`T->addr`, and the only consumer of `T->sas`.  It never runs, so the strduped
address and the whole resolved `struct sock_addr **` list are both leaked.

**2. The caller is never called back.**  `dnsthread_resolveone()` returns -1,
so the caller reasonably concludes nothing is pending — but the worker
completes the resolution and writes its completion byte anyway.

**3. The stale byte makes the next resolution free memory in use.**  The
worker sets `T->state` back to `THREAD_SLEEPING`, so a later
`dnsthread_resolveone()` on the same `T` proceeds normally.  This time the
registration succeeds — and `wakeupsock[1]` is *already* readable, because the
byte from the abandoned resolution was never drained.  So:

* the worker takes `addr = T->addr` under the mutex, releases the mutex, and
  calls `sock_resolve(addr)`;
* the event loop fires `callback_resolveone()` for the stale byte, which locks
  the mutex and does `free(T->addr)` — the same buffer the worker is passing
  to `sock_resolve()` right now.

That is a use-after-free across the two threads.  The mutex does not prevent
it: the worker deliberately drops the mutex before resolving, precisely
because resolution is slow.

The same callback then hands `T->sas` — the *previous* resolution's result —
to the new request's callback, and the result the worker eventually stores is
leaked in turn.

### Reachability

`events_network_register()` fails on allocation failure, which is the same
reachability as #427 / tarsnap#710.

### Fix

PR to follow: register the wakeup listener *before* setting `THREAD_HASWORK`
and signalling the worker, so a registration failure leaves the thread
untouched and the address can simply be freed.  The `pthread_cond_signal()`
failure path gets a matching unwind.

### How this was found

Source review of `lib/dnsthread/dnsthread.c` at current master, checking each
early return in `dnsthread_resolveone()` against the state the worker thread
had already been handed.

woahwhattheheck:

Fixes #438.

`dnsthread_resolveone()` set `THREAD_HASWORK` and signalled the worker thread
*before* registering the wakeup listener, and `err1` only unlocked the mutex.
A failed `events_network_register()` therefore returned -1 with a resolution
already running and nothing registered to read its completion byte: `T->addr`
and `T->sas` both leaked, the caller's callback never fired, and the
undrained byte made the next `dnsthread_resolveone()` fire
`callback_resolveone()` immediately — freeing `T->addr` while the worker
thread was passing that same pointer to `sock_resolve()`.

### The change

Register the listener before handing the work over, and unwind properly:

* `err2` frees the strduped address — reached when the registration fails,
  at which point the worker has not been touched.
* `err3` additionally restores `THREAD_SLEEPING` and cancels the
  registration — reached only if `pthread_cond_signal()` fails.
* `err1` is unchanged and still handles the `strdup()` failure, where there is
  nothing to free.

`events_network_cancel()`'s return is ignored, matching `network_accept.c`,
`network_connect.c`, `network_read.c` and `network_write.c`.

### Note

I did not try to make the failure path take the work *back* from the worker
once it is running, because it cannot be done safely — the worker drops the
mutex for the duration of `sock_resolve()`.  Doing the registration first
means that situation never arises.

### DIFF
--- a/lib/dnsthread/dnsthread.c
+++ b/lib/dnsthread/dnsthread.c
@@ -265,20 +265,25 @@ dnsthread_resolveone(DNSTHREAD T, const char * addr,
 	T->callback = callback;
 	T->cookie = cookie;
 
+	/*
+	 * We want a callback when the worker thread pokes us.  Register this
+	 * before handing the work over: once the thread is running we cannot
+	 * take the work back, and without this registration nothing would
+	 * ever read the completion byte or free the results.
+	 */
+	if (events_network_register(callback_resolveone, T, T->wakeupsock[1],
+	    EVENTS_NETWORK_OP_READ)) {
+		warnp("Error registering wakeup listener");
+		goto err2;
+	}
+
 	/* There is now work for the thread to do. */
 	T->state = THREAD_HASWORK;
 
 	/* Wake up the worker thread. */
 	if ((rc = pthread_cond_signal(&T->cv)) != 0) {
 		warn0("pthread_cond_signal: %s", strerror(rc));
-		goto err1;
-	}
-
-	/* We want a callback when the worker thread pokes us. */
-	if (events_network_register(callback_resolveone, T, T->wakeupsock[1],
-	    EVENTS_NETWORK_OP_READ)) {
-		warnp("Error registering wakeup listener");
-		goto err1;
+		goto err3;
 	}
 
 ealready:
@@ -295,6 +300,11 @@ ealready:
 	/* Success! */
 	return (0);
 
+err3:
+	T->state = THREAD_SLEEPING;
+	events_network_cancel(T->wakeupsock[1], EVENTS_NETWORK_OP_READ);
+err2:
+	free(T->addr);
 err1:
 	pthread_mutex_unlock(&T->mtx);
 err0:


tokenjunkielabs:
Following up on first-reporter bug bounty #438. This PR is the `woahwhattheheck` carrier for the DNS worker/listener ordering race which can abandon a running resolution. Please review the report/fix when convenient and classify #438 under the repository's published bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #441, opened 2026-09-06
proto_handshake: clear the cookie before freeing it

### LINKED ISSUE #440: [bug bounty] The handshake cookie, holding the diffie-hellman private value, is freed without being wiped

### Summary

`struct handshake_cookie` holds `x`, the diffie-hellman private value, along
with the two diffie-hellman parameter MAC keys.  It is `free()`d at four
places in `lib/proto/proto_handshake.c` and never wiped at any of them, so
that material is left in freed heap memory after every connection.

Same class as #426 / #427 and Tarsnap/tarsnap#709, #736 and #743, in a file
none of those touched.

### The material

```c
struct handshake_cookie {
...
	uint8_t nonce_local[PCRYPT_NONCE_LEN];
	uint8_t nonce_remote[PCRYPT_NONCE_LEN];
	uint8_t dhmac_local[PCRYPT_DHMAC_LEN];
	uint8_t dhmac_remote[PCRYPT_DHMAC_LEN];
	uint8_t x[PCRYPT_X_LEN];
	uint8_t yh_local[PCRYPT_YH_LEN];
	uint8_t yh_remote[PCRYPT_YH_LEN];
...
};
```

`PCRYPT_X_LEN` is `CRYPTO_DH_PRIVLEN`: `x` is the private exponent generated
by `proto_crypt_dh_generate()` and consumed by `proto_crypt_mkkeys()`.  Its
whole purpose is to be forgotten once the session keys exist — that is what
makes the `-f`/PFS handshake forward-secret.  `dhmac_local` and `dhmac_remote`
are the keys `proto_crypt_dhmac()` derives from the shared secret.

The nonces and the `yh_*` values are sent over the wire in the clear and are
not the concern here; `x` and the `dhmac_*` pair are.

### The four free sites

None of them wipe:

* `handshakefail()` — after the failure callback.
* `proto_handshake()` at `err1` — the allocation-and-setup error path.
* `handshakedone()` — the normal completion path, after the keys have been
  handed to the caller.  This is the one that runs for every successful
  connection.
* `proto_handshake_cancel()` — when the connection is dropped mid-handshake.

`handshakedone()` is worth calling out: it runs once per successful
connection, and it is exactly the point at which `x` has served its purpose
and should stop existing.

Nothing else in the file zeroes the cookie, and `insecure_memzero.h` is not
currently included.

### Fix

PR to follow: `insecure_memzero(H, sizeof(struct handshake_cookie))` before
each of the four `free(H)` calls, and the corresponding include.

I wipe on all four rather than only where `x` is known to be populated, to
match `scryptenc.c`, which wipes `dk` at `err1` on paths where it may never
have been filled in.

### How this was found

Source review of `lib/proto/proto_handshake.c` at current master, pairing the
sensitive members of the cookie against each site which frees it.

woahwhattheheck:

Fixes #440.

`struct handshake_cookie` holds `x` — the diffie-hellman private value, whose
being forgotten is what makes the PFS handshake forward-secret — and the two
diffie-hellman parameter MAC keys derived from the shared secret.  All four
sites which free the cookie did so without wiping it:

* `handshakefail()`
* `proto_handshake()` at `err1`
* `handshakedone()`, which runs once per successful connection, at exactly the
  point where `x` has served its purpose
* `proto_handshake_cancel()`

Four `insecure_memzero()` calls and the include; no other change.

### Why all four rather than only where `x` is populated

`err1` is reached before the diffie-hellman value is ever generated, so
strictly it has nothing to protect.  I wiped it anyway so that "the cookie is
always cleared before it is freed" is a property that holds without the reader
having to trace which paths populate what — the same reasoning as
`scryptenc.c`, which wipes `dk` at `err1` on paths where it may never have
been filled in.  Happy to drop that one if you would rather keep it exact.

The nonces and `yh_*` values in the cookie go over the wire in the clear, so
they are not the point here; they are simply covered by wiping the whole
structure, which is cheaper and less fragile than picking members.

### DIFF
--- a/lib/proto/proto_handshake.c
+++ b/lib/proto/proto_handshake.c
@@ -4,6 +4,7 @@
 #include <unistd.h>
 
 #include "crypto_entropy.h"
+#include "insecure_memzero.h"
 #include "network.h"
 
 #include "proto_crypt.h"
@@ -53,7 +54,8 @@ handshakefail(struct handshake_cookie * H)
 	/* Perform the callback. */
 	rc = (H->callback)(H->cookie, NULL, NULL);
 
-	/* Free the cookie. */
+	/* Clear sensitive material and free the cookie. */
+	insecure_memzero(H, sizeof(struct handshake_cookie));
 	free(H);
 
 	/* Return status from callback. */
@@ -112,6 +114,7 @@ proto_handshake(int s, int decr, int nopfs, int requirepfs,
 err2:
 	network_write_cancel(H->write_cookie);
 err1:
+	insecure_memzero(H, sizeof(struct handshake_cookie));
 	free(H);
 err0:
 	/* Failure! */
@@ -298,7 +301,8 @@ handshakedone(struct handshake_cookie * H)
 	/* Perform the callback. */
 	rc = (H->callback)(H->cookie, c, s);
 
-	/* Free the cookie. */
+	/* Clear sensitive material and free the cookie. */
+	insecure_memzero(H, sizeof(struct handshake_cookie));
 	free(H);
 
 	/* Return status code from callback. */
@@ -326,6 +330,7 @@ proto_handshake_cancel(void * cookie)
 	if (H->write_cookie != NULL)
 		network_write_cancel(H->write_cookie);
 
-	/* Free the cookie. */
+	/* Clear sensitive material and free the cookie. */
+	insecure_memzero(H, sizeof(struct handshake_cookie));
 	free(H);
 }


tokenjunkielabs:
Following up on first-reporter bug bounty #440. This PR is the `woahwhattheheck` carrier for the unwiped handshake cookie containing Diffie-Hellman private/key material. Please review the report/fix when convenient and classify #440 under the repository's published bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #448, opened 2026-09-09
graceful_shutdown: clarifications and safety checks

gperciva:


lib/util/graceful_shutdown.c
```diff
@@ -96,34 +110,46 @@ graceful_shutdown_initialize(int (* begin_shutdown_parent)(void *),
...
  */
-void
+int
 graceful_shutdown_manual(void)
 {
 
 	/* Sanity check: we must be initialized. */
-	assert(begin_shutdown != NULL);
-	assert(caller_cookie != NULL);
+	assert(initialized);
```
gperciva, line 148: There's a decent chance that you'll consider `initialized` to be overkill; as commit 3 notes, it's obvious that `_initialize()` must complete successfully before one can call `_manual()`.  If so, I'm quite content to remove that commit.

spipe/main.c
```diff
@@ -54,7 +54,7 @@ callback_conndied(void * cookie, int reason)
 
 	/* Shut it down if there's an error. */
 	if (ET->connection_error)
-		graceful_shutdown_manual();
+		(void)graceful_shutdown_manual();
```
gperciva, line 57: This doesn't fix the problem of discarding the return value, but that's outside of the scope of this PR.  Another PR is coming soon which will fix issues in `spipe/main.c`.

gperciva:
On second thought, I don't think that 3rd commit is actually doing anything that `assert(begin_shutdown != NULL);` didn't cover.

I'll re-check and bring this PR back (possibly modified) later.

gperciva:
Ignore that previous (now-hidden) comment; I'd forgotten that there was a difference between "began _initialize()" and "completed _initialize()".

